PR #22, opened 2024-11-23
Add documentation on how to use locally built version

cypressious:


Contribute.md
```diff
@@ -32,4 +32,29 @@ participate in the orchestration by listening for messages and broadcasting UI r
 
 When the dev runtime is launched, a Gradle Daemon is launched in continuous mode, which will connect
 to the orchestration and send 'ReloadClassesRequests' when the actual .class files for the runtime classpath
-changes.
\ No newline at end of file
+changes.
+
+# Use locally built version in your project
```
sellmair, line 37: Using `publishToMavenLocal` would be an alternative. Would you also be happy with this approach?

sellmair (approved):

Contribute.md
sellmair: Ah you know what: mavenLocal is lazy and error prone. Its better to go with your approach and adverstise it this way. Lets go to master.

cypressious: Sure, would you like me to create another PR?

PR #29, opened 2024-11-28
Update README.md

SebastianAigner:
Removes outdated information that may be confusing for people who want to try CHR.


sellmair:
Thank you 😃

PR #32, opened 2024-11-30
Comply to ClassFileTransformer contract in ComposeGroupInvalidationKe…

qwwdfsad:
…yTransformer

Return 'null' for non-transformed classes, otherwise, actual transformation happens and additional classloading work/RSS bookkeeping (for retransformations) is done


sellmair (approved):
Thank you 😍

PR #48, opened 2024-12-13
Allow setting up server port from build tool

singleton11:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=1.0.0-dev.31.6
+version=1.0.0-dev.31.7
```
sellmair: Probably a left over from local testing?

sellmair (changes requested):

sellmair:
Out of curiosity: Why would you need a fixed port, known by the build tool?

gradle.properties
singleton11: Did it on purpose, thought that it's a good idea to bump version
Gonna remove it though to stick with the default release process

PR #54, opened 2024-12-16
Add Amper support

singleton11:


hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/recompiler.kt
```diff
@@ -35,34 +41,61 @@ private val recompileRequests = LinkedBlockingQueue<RecompileRequest>(
...
-    val gradleBuildTask = gradleBuildTask ?: run {
-        logger.error("Missing 'compose.build.compile.task' property")
-        return
-    }
 
     val port = ComposeHotReloadAgent.orchestration.port
     logger.debug("'Compose Recompiler': Using orchestration at '$port'")
 
+    val processBuilder = when (buildSystem ?: "gradle") {
+        "amper" -> {
```
sellmair: What do you think about modelling this as enum, so that the name of the enum can be easily used as param?

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/recompiler.kt
```diff
@@ -35,34 +41,61 @@ private val recompileRequests = LinkedBlockingQueue<RecompileRequest>(
...
+            createRecompilerProcessBuilder(
+                gradleBuildRoot = composeBuildRoot,
+                gradleBuildProject = gradleBuildProject,
+                gradleBuildTask = gradleBuildTask,
+                orchestrationPort = port
+            )
+        }
+
+        else -> {
+            logger.error("${HotReloadProperty.BuildSystem.key} has incorrect value: $buildSystem")
```
sellmair: Maybe use logger.orchestration here, to send this to other tools, as this would be a very important message which should also be displayed in the IDE or dev tooling!

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/environment.kt
```diff
@@ -11,13 +11,17 @@ public enum class HotReloadProperty(public val key: String) {
     IsHeadless("compose.reload.headless"),
     HotClasspath("compose.reload.hotApplicationClasspath"),
 
+    BuildSystem("compose.reload.build.system"),
```
sellmair: NIT: `compose.reload.buildSystem` would be consistent, wdyt?

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/recompiler.kt
singleton11: It depends on who we think the consumer of these messages will be, I don't want to show it as anything user-visible in the IDE for instance, it's more like invariant according to me (that implies of course that other subsystems, like build tool passes the proper value here, however, I don't expect versioning of this thing, because for our purposes, it's easier to break something, if needed)

As debug information it would be useful though, and for that purposes, error logging is enough, I guess

I don't consider build system authors as consumers of these messages because I don't expect more than 2 of these we have right now in foreseeable future

sellmair (approved):

PR #59, opened 2025-01-10
Improve statics re-initialization

sellmair:


hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/Try.kt
```diff
@@ -0,0 +1,52 @@
...
+import kotlin.contracts.contract
+import kotlin.coroutines.cancellation.CancellationException
+
+public typealias Try<T> = Either<T, Throwable>
+
+public inline fun <T> Try(block: () -> T): Try<T> {
+    return try {
+        block().toLeft()
+    } catch (t: Throwable) {
+        if (t is CancellationException) throw t
```
brokenhappy, line 13: Sorry, you didn't ask, but I got distracted by this lil nitpick, while indeed mostly what you need. There are cases where this might start throwing rogue cancellation exceptions. This becomes increasingly less common in properly structured concurrency (code that does not pass around `CoroutineScope`s and/or `Job`s).
But if you want to be extra safe, you should use `ensureActive()` here instead.
I usually refer to this read:
https://betterprogramming.pub/the-silent-killer-thats-crashing-your-coroutines-9171d1e8f79b

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/topologicalSort.kt
```diff
@@ -0,0 +1,193 @@
...
+    return ranked.toMutableSet()
+}
+
+private fun <T> Set<T>.buildNodes(
+    edges: (T) -> Iterable<T>
+): List<Node<T>> {
+    if (this.isEmpty()) return emptyList()
+    var index = 0
+
+    val nodes = mapTo(ArrayList(size * 2)) { value -> Node(value, index++) }
```
brokenhappy: Again, not the feedback you asked for 😅, but `mapIndexedTo` exists :)

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/topologicalSort.kt
```diff
@@ -0,0 +1,193 @@
...
+    val nodesMap = nodes.associateByTo(LinkedHashMap(nodes.size)) { it.value }
```
brokenhappy: Does this need to be linked?

sellmair: No, but interestingly, the benchmark prefers LinkedHashMap here 🤷 That was a surprise to me.

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/Try.kt
sellmair: That's a good catch. I skimmed over the article and will read it more carefully today evening. 
However, since this `Try` is not a suspend function, I hope that this should just act as 'thin wall' and the rogue cancellation will be handled similar to the `Try` being absent.

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/topologicalSort.kt
sellmair: Done, thank you ☺️

PR #73, opened 2025-02-05
Setup Apache2 license

sellmair:


LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2025 JetBrains s.r.o
```
erokhins: and contributors

buildSrc/build.gradle.kts
```diff
@@ -1,3 +1,8 @@
+/*
```
erokhins, line 1: seems like for build files copyright header is optional (my feeling based on kotlin repo)

erokhins (changes requested):
Please add "and Contributors" to the LICENSE file, the rest is good

buildSrc/build.gradle.kts
sellmair: Doesn't harm and IntelliJ's default configuration decided to add the copyright, so probably its the easiest to keep it default 🙇

PR #88, opened 2025-02-13
add local.properties to .gitignore

hamedmonji:
removes local.properties from git and add it to .gitignore


sellmair:
Thank you!

PR #105, opened 2025-03-08
Don't show icon of the devtools window in the taskbar

natanfudge:
`Window` was replaced with `DialogWindow` which uses `ComposeDialog`, see https://github.com/JetBrains/compose-multiplatform-core/pull/177 , which uses `JDialog` instead of `JFrame`, see https://stackoverflow.com/questions/32703918/hide-window-from-task-bar/50314627#50314627.

The logo is now unused so I removed it. 

Everything else works the same as before. 

Also see https://github.com/JetBrains/compose-hot-reload/discussions/91#discussioncomment-12435280


sellmair:
Oh Jees! It would have been that easy? Thank you!! 🙇  
Testing the code and then sending it to staging

sellmair:
Works fine on macOS and windows 👍 
Thank you very much! I thought it's much more complicated than this. Very happy for this contribution. 
Sent to staging. Shall land in dev builds soon. Expected in alpha02

natanfudge:
Thank *you* for building this amazing tool :)

PR #106, opened 2025-03-11
Fix sidecar window failing to expand on Linux

Nohus:
On Linux, the tooling window fails to expand when the icon is clicked, until it's next recomposed with a new `WindowState`. This PR forces the update which fixes the problem. The change is only applied on Linux.

I am not sure what's the underlying reason for this being required (or not required on other systems), there might be a better solution if the reason is found.

### DIFF
--- a/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/OperatingSystem.kt
+++ b/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/OperatingSystem.kt
@@ -1,0 +1,10 @@
+/*
+ * Copyright 2024-2025 JetBrains s.r.o. and Compose Hot Reload contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.compose.reload.jvm.tooling
+
+val isLinux by lazy {
+    "linux" in System.getProperty("os.name").lowercase()
+}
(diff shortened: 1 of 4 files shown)


sellmair:
Thank you very much for this PR!
I will check out your branch soon and try to understand what the underlying issue is. 
The code in itself looks fine 👍

sellmair:
Can you check if the following commit also works for you?
b127ded774cda4f276fda8319b34d6e45fd5499d

This commit will also be shipped in the next dev builds

sellmair:
If the commit works, then I think we could close this MR. The investigation from your side was very helpful in any case!

Nohus:
That commit doesn't seem to fix the issue, when clicking the button it moves but doesn't expand until the application window is moved.

sellmair:
This means that the only solution, as of right now, would be what you came up with: 
The composable sends a message where the window state gets re-emitted?
I would really want to know what the underlying root cause would be here.
I tested on a remote linux box and I was able to reproduce the issue we have seen, but the commit mentioned here fixed it visiually there.

sellmair:
I have spent some time now on that issue and its strange because the DialogState passed to the dialog has the correct sizes in it. There is some code which actively writes old values to this state. 

I found another way of reproducing this and found that reading the state value (which listens for the rogue change) will work. 
There was a commit pushed to staging (available in master soon, then in the next dev builds). Maybe you could check that one out at some point?

Nohus:
I can confirm it's fixed on the `staging` branch. :+1:

Nohus:
I saw you replaced `private val windowGap = 12.dp` with hardcoded `12.dp` in both `getSideCarWindowSize` and `animateWindowPosition`, not sure if that's intended, they shouldn't ever get out of sync.

sellmair:
I'll close the MR as it seems like the underlying issue is still in compose desktop itself. 
Regarding the hardcoded padding and spacing values. The devtools is currently more a 'proof of concept'. 
The entire project will focus on stabilizing the core tech first (compiler + runtime + agent). Once this is done, I hope that another engineer, who is much better at creating applications than me, will take over and build it properly!

PR #108, opened 2025-03-11
Removed window transparency on Linux to workaround skiko bug

Nohus:
On Linux, transparent windows cannot be used as they break hardware accelerated rendering due to a bug in skiko: https://youtrack.jetbrains.com/issue/CMP-5957/Transparent-windows-break-hardware-acceleration-on-Linux

This PR disables window transparency on the sidecar window on Linux.
- Sidecar window no longer has rounded corners
- The expanding/collapsing animation is removed, as it cannot work without the window being transparent
- The `DtReloadStatusBanner` is removed, as it also relies on the transparency. The functionality of seeing the reload status is still available, as even in the collapsed state it's visible on the icon.

These changes only apply on Linux. When running on macOS and Windows, nothing is affected. This does unfortunately make the tooling window look less attractive, ideally the changes can be reverted when the skiko bug is fixed.

### DIFF
--- a/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
+++ b/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
@@ -7,7 +7,9 @@ package org.jetbrains.compose.reload.jvm.tooling.sidecar
 
 import androidx.compose.animation.AnimatedContent
 import androidx.compose.animation.core.Animatable
-import androidx.compose.animation.core.animateDpAsState
+import androidx.compose.animation.core.FiniteAnimationSpec
+import androidx.compose.animation.core.SnapSpec
+import androidx.compose.animation.core.SpringSpec
 import androidx.compose.animation.core.tween
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
@@ -15,9 +17,9 @@ import androidx.compose.animation.scaleIn
 import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
@@ -32,6 +34,7 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.DpSize
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.window.DialogState
 import androidx.compose.ui.window.DialogWindow
@@ -41,6 +44,7 @@ import kotlinx.coroutines.delay
 import org.jetbrains.compose.reload.core.WindowId
 import org.jetbrains.compose.reload.core.createLogger
 import org.jetbrains.compose.reload.jvm.tooling.invokeWhenMessageReceived
+import org.jetbrains.compose.reload.jvm.tooling.isLinux
 import org.jetbrains.compose.reload.jvm.tooling.orchestration
 import org.jetbrains.compose.reload.jvm.tooling.theme.DtColors
 import org.jetbrains.compose.reload.jvm.tooling.widgets.DtButton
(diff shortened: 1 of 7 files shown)


hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
```diff
@@ -58,6 +61,10 @@ import kotlin.time.Duration.Companion.milliseconds
 private val logger = createLogger()
 
 private val DevToolingSidecarShape = RoundedCornerShape(8.dp)
+private val DtButtonSize = 28.dp
+private val windowGap = 12.dp // Space between the application window and sidecar window
+private val isUsingTransparentWindow = !isLinux
```
sellmair: I think this 'isLinux' property is missing in this commit, right? 🤔

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
```diff
@@ -89,23 +95,23 @@ fun DtSidecarWindow(
     val x by xAnimatable.asState()
     val y by yAnimatable.asState()
 
-    val height by animateDpAsState(targetValue = maxOf(windowState.size.height, 512.dp))
+    val animationSpec: FiniteAnimationSpec<Float> = if (isSizeAnimationEnabled) SpringSpec() else SnapSpec()
 
     LaunchedEffect(windowState.position) {
-        xAnimatable.animateTo(targetX.value)
+        xAnimatable.animateTo(targetX.value, animationSpec)
```
sellmair: Is the animation on the window position intentionally removed?

sellmair:
I am unsure about the removal of the banner. Some more dedicated visual representation would really be nice. Maybe even if the bar is just very small in height?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
```diff
@@ -198,3 +226,50 @@ fun DtSidecarWindowContent(
...
+                .animatedReloadStatusBorder(
+                    shape = DevToolingSidecarShape,
+                    idleColor = if (isExpanded) Color.LightGray else Color.Transparent
+                )
+                .clip(DevToolingSidecarShape)
+                .background(DtColors.applicationBackground)
+                .weight(1f, fill = false),
+            contentAlignment = Alignment.TopEnd,
+        ) {
+            if (!isExpanded) {
```
sellmair: Have you thought about other approaches to disabling the animation? I see that this duplicates code a little. 
The issue is not the code duplication, but the fact that the code will be duplicated for host platforms. I might see myself easily maintaining this code, testing on macos and then forgetting to test on linux as well leading to inconsistencies which we would have to rely on manual QA to find 🤔

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
Nohus: You are correct, this commit was on top of my other PR, but I rebased it when making this PR and clearly forgot a file. Fixed.

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
Nohus: Yes, since the resizing animation is removed, animation the position must also be removed, otherwise the tooling window would instantly cover the application window, then move away.

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
Nohus: I have inititally modified the `DtSidecarWindowContent` function to disable/enable animations as needed, but I saw two issues with that:
- It was making the function complex, plus having an `AnimatedContent` that sometimes doesn't animate didn't feel right, with `EnterAnimation.None.togetherWith(ExitAnimation.None)`
- I felt having a separate method makes it easier to remove later once the workaround is not longer required

But if you think it's better not to have the duplicate, I am happy to do it the other way.

Nohus:
Readded the missing `OperatingSystem.kt` that I accidentally didn't include in the commit.

Nohus:
> I am unsure about the removal of the banner. Some more dedicated visual representation would really be nice. Maybe even if the bar is just very small in height?

The icon itself animations in the same way, giving that representation, or do you think it's not enough?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtSidecarWindow.kt
sellmair: But if the target window just gets moved around, then the position animation should still be correct, or am I missing something?

Nohus: You are absolutely correct, `snapTo` is already used in the case I was thinking of, so disabling this animation is not required. Updated.

sellmair:
Thank you very much! Should be available in the next dev build 🙇

PR #139, opened 2025-04-17
Extract interfaces for Gradle model entities

avan1235:
Extracts interfaces for Gradle model entities to avoid exceptions like
```
com.intellij.openapi.externalSystem.model.ExternalSystemException: org.jetbrains.compose.reload.gradle.idea.IdeaComposeHotReloadModel is not an interface
```
and suggest some improvements in serialisation stragegy


sellmair:
Thank you 🙇

PR #143, opened 2025-04-22
Add copy functions to Gradle IDEA model

avan1235:


sellmair:
I fixed the apiDump and pushed to staging 🚀 
Thank you 🙇

Issue #152, opened 2025-04-25
`parseRuntimeInstructionTree` works incorrectly with eager returns


AbdullinAM:
# Reproducing example

* Put this code in `/samples/counter/app/src/jvmMain/kotlin/App.kt`:

```kotlin
import androidx.compose.foundation.layout.*
import androidx.compose.material.Text
import androidx.compose.runtime.*

@Composable
fun Value(): String? {
    return "value: 0"
}

@Composable
fun App() {
    val value = Value() ?: return
    Column {
        Text(value)
    }
}
```

* Run the `counter` app

* Change `return "value: 0"` in `Value` function to `return "value: 1"`

* Save the code and trigger a reload

* See, that the app still shows the text `"value: 0"`

Running the same scenario with `val value = Value()!!` instead of eager return works correctly.

# Cause

`ClassNode.renderRuntimeInstructionTree()` generates the following `RuntimeInstructionTree` for this code:
```
    fun Foo (Landroidx/compose/runtime/Composer;I)V {
        Method (group=ComposeGroupKey(key=-607362160)) [0:30] {
            LabelToken(L0)
            BlockToken(2)
            BlockToken(3)
            JumpToken(L5, opocde=167)
            LabelToken(L4)
            BlockToken(2)
            LabelToken(L5)
            BlockToken(1)
            ReturnToken(returnInsn=177)
    
            RestartGroup (group=ComposeGroupKey(key=-607362160)) [3:21] {
                StartRestartGroup(key=ComposeGroupKey(key=-607362160))
                BlockToken(2)
                SourceInformation(C(Foo)11@231L7,12@253L57:Test.kt)
                BlockToken(1)
                JumpToken(L1, opocde=154)
                BlockToken(2)
                JumpToken(L78, opocde=154)
                LabelToken(L1)
                BlockToken(2)
                JumpToken(L2, opocde=153)
                BlockToken(5)
                LabelToken(L2)
                BlockToken(6)
                JumpToken(L6, opocde=199)
                BlockToken(2)
                JumpToken(L3, opocde=153)
                BlockToken(1)
                LabelToken(L3)
                BlockToken(2)
                EndRestartGroup()
            }
        }
    }
```

This instruction tree contains labels `L0 -- L5`. However, analysing the bytecode of this function we will see that it contains labels `L0 -- L89`, so the majority of the bytecode was not parsed.

Tokenization works correctly, so the issue is with parsing.

# Investigation

[...]

PR #157, opened 2025-04-30
Disable source information compiler flag in tests

AbdullinAM:


hot-reload-analysis/src/test/resources/runtimeInstructionTree/org.jetbrains.compose.reload.analysis.tests.RuntimeInstructionTreeParserTest/test - #152 - composable with eager return/runtime-instructions-tree.txt
```diff
@@ -193,47 +191,44 @@ class TestKt {
...
                 LabelToken(L60)
-                BlockToken(1)
-                SourceInformationMarkerStart(key=ComposeGroupKey(key=-333101486), sourceInformation='C13@270L11:Test.kt')
                 BlockToken(23)
-                SourceInformationMarkerEnd
+                LabelToken(L61)
+                BlockToken(2)
                 LabelToken(L62)
                 BlockToken(2)
+                SourceInformationMarkerEnd
```
sellmair, line 199: Wait, I see that the start is removed, but why is there still a 'SourceInformationMarkerEnd'?

AbdullinAM: In this test, the source information markers are added in the process of inlining `Column` function call. We cannot remove these markers markers, because `sourceInformation` flag does not affect them. We only remove these markers from **our** code. 

And all the `SourceInformationMarkerStart`/`SourceInformationMarkerEnd` markers are still consistent (i.e. for each start there is an end)

Issue #123, opened 2025-03-31
Potentially allow 'return' w/o calls to end(Restart/Replace)Group


sellmair:
```
WARN [org.jetbrains.compose.reload.analysis.RuntimeInstructionTreeKt]: 'parseRuntimeInstructionTree' failed on com/jetbrains/toolbox/ui/ToolboxUi$MainScreenBody$1.invoke (Landroidx/compose/runtime/Composer;I)V: Failure(message=ReturnToken is not allowed in ReplaceGroup scope, throwable=null)
```

PR #158, opened 2025-04-30
Add source renders to tree parsing tests

AbdullinAM:


hot-reload-analysis/src/test/resources/runtimeInstructionTree/org.jetbrains.compose.reload.analysis.tests.RuntimeInstructionTreeParserTest/test - #123 - composable with local return/runtime-instructions-source.txt
```diff
@@ -0,0 +1,24 @@
+import androidx.compose.runtime.*
+import androidx.compose.material3.Text
+ 
+@Composable
+fun Foo(value: Int) {
+    Text("A")                                               	// Method (group=ComposeGroupKey(key=-643616831))
+    if(value > -12) {                                       	// RestartGroup (group=ComposeGroupKey(key=-643616831))
+        Bar {                                               	// RestartGroup (group=ComposeGroupKey(key=-643616831)),ReplaceGroup (group=ComposeGroupKey(key=-1254930869))
```
sellmair: I would be interested to learn why this line still belongs technically to `RestartGroup (group=ComposeGroupKey(key=-643616831))` as well?

AbdullinAM: This was because of 'off by 1' error with line indices 😅

PR #160, opened 2025-05-02
Option to always show the reload counter in dev tools

AbdullinAM:
Simple implementation that allows to always show the reload counter. Waiting for feedback on UI   😃. Also, I'm not sure if it should be enabled by default.

### DIFF
--- a/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtReloadCounterStatusItem.kt
+++ b/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtReloadCounterStatusItem.kt
@@ -5,10 +5,12 @@
 
 package org.jetbrains.compose.reload.jvm.tooling.sidecar
 
+import androidx.compose.foundation.layout.Row
 import androidx.compose.material.Icon
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Refresh
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.scale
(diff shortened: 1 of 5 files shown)


sellmair:
The reload logo might be a little small. Is this on my machine?
Other than that, I would say let’s enable it by default, this way we could also think of getting rid of the property.

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/reload/jvm/tooling/sidecar/DtReloadCounterStatusItem.kt
```diff
@@ -38,12 +40,18 @@ fun DtCollapsedReloadCounterStatusItem() {
...
+        state.successfulReloads < 10 -> 1.0f
+        state.successfulReloads < 100 -> 0.9f
+        else -> 0.7f
+    }
+    Row(
+        verticalAlignment = Alignment.CenterVertically,
+        modifier = Modifier.scale(textScale)
+    ) {
+        if (state.successfulReloads < 100) {
+            DtText("⟳", modifier = Modifier.scale(1.25f))
```
sellmair: 😬 I think you should use the same icon as in `DtReloadCounterStatusItem`

AbdullinAM: Switched to the default icon. I had to scale down the text a little bit to make it fit (this is why I originally opted to use the unicode character). But I think it is still readable.

sellmair: I would suggest using the size modifier in this case, but I am ok for landing this

sellmair (approved):

PR #165, opened 2025-05-07
CLI reload facilities: Allow reloading using a generic ./gradlew reload command

sellmair:


hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
```diff
@@ -53,6 +53,7 @@ public sealed class OrchestrationMessage : Serializable {
      *
      * Note: This message can be sent multiple times!
      */
+    @Deprecated("This message is not required anymore")
```
AbdullinAM, line 61: Do we need to keep this class if it is not used anymore? If yes, maybe we should add a small comment on wty it's still here and why its deprecated?

AbdullinAM (commented):
Really great changes! 

One thing: `GradleRecompilerProcessTest.test - gradle recompiler process is stopped` test is failing with timeout currently. Does this reproduce for you as well?

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
sellmair: We already try keeping binary compatibility for clients such as IntelliJ or Amper. I'll add a comment!

sellmair:
> Really great changes!
> 
> One thing: `GradleRecompilerProcessTest.test - gradle recompiler process is stopped` test is failing with timeout currently. Does this reproduce for you as well?

Thank you! I cannot see the failure on CI or on my machine (even when using @Deflake) 🤔

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
sellmair: I added a comment about the deprecation.

PR #167, opened 2025-05-08
Restore gradient brush animation to reload bar

zsmb13:
https://github.com/user-attachments/assets/35b5b46e-a54f-4f96-8829-13c780c31c03


sellmair:
Thank you 🙇

Issue #178, opened 2025-05-12
Changes from Non-Composable entry point are not invalidated


sellmair:
The simplest code example is

From: 
```kotlin
fun main() {
    /* Non Composable World */
    singleWindowApplication { 
        /* Composable World */
        Text("Hello")
    }
```

To:
```kotlin
fun main() {
    /* Non Composable World */
   var myVariable = 0

    singleWindowApplication {
        /* Composable World */
        myVariable = 1 // <- Capture will change lambda passed 'into the Composable World'
        Text("Hello: $myVariable")
    }
```

The change will not be reflected, the old lambda can cause all sorts of issues (see #170)

Issue #170, opened 2025-05-08
NoSuchMethodError on reload when making specific changes


natanfudge:
1. 
Use this code:

```kotlin
fun main() {
//    val list = mutableListOf<@Composable () -> Unit>()
//    list.add {
//
//    }
    application {
        Window(::exitApplication) {
            Text("gg")
//            list.forEach { it() }
        }
    }
}
```
2. `gradlew runHot`

3. Add back all commented lines
4. Save
5. Observe: Crash:
```
Exception in thread "main" java.lang.NoSuchMethodError: 'kotlin.jvm.functions.Function3 ComposableSingletons$MainKt.getLambda$-1642765418$MyProject()'
        at ComposableSingletons$MainKt$lambda$1632412612$1.invoke(Main.kt:19)
        at ComposableSingletons$MainKt$lambda$1632412612$1.invoke(Main.kt:18)
        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jb.kt:116)
        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jb.kt:33)
        at androidx.compose.ui.window.Application_desktopKt$application$1$1.invoke(Application.desktop.kt:117)
        at androidx.compose.ui.window.Application_desktopKt$application$1$1.invoke(Application.desktop.kt:116)
        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jb.kt:116)
        at androidx.compose.runtime.internal.ComposableLambdaImpl$invoke$2.invoke(ComposableLambda.jb.kt:128)
        at androidx.compose.runtime.internal.ComposableLambdaImpl$invoke$2.invoke(ComposableLambda.jb.kt:127)
        at androidx.compose.runtime.RecomposeScopeImpl.compose(RecomposeScopeImpl.kt:192)
        at androidx.compose.runtime.ComposerImpl.recomposeToGroupEnd(Composer.kt:2825)
        at androidx.compose.runtime.ComposerImpl.skipCurrentGroup(Composer.kt:3116)
        at androidx.compose.runtime.ComposerImpl.doCompose(Composer.kt:3607)
        at androidx.compose.runtime.ComposerImpl.recompose$runtime(Composer.kt:3552)
        at androidx.compose.runtime.CompositionImpl.recompose(Composition.kt:948)
[...]

Issue #169, opened 2025-05-08
When reload fails, runHot does not exit


natanfudge:
1. 
Use this code:

```kotlin
fun main() {
//    val list = mutableListOf<@Composable () -> Unit>()
//    list.add {
//
//    }
    application {
        Window(::exitApplication) {
            Text("gg")
//            list.forEach { it() }
        }
    }
}
```
2. `gradlew runHot`

3. Add back all commented lines
4. Save
5. Observe: Crash: `java.lang.NoSuchMethodError`, window closes but runHot keeps running and i need to terminate it manually.

using CHR alpha-09, Windows 11, Gradle

sellmair:
This is because crashes are supposed to be recoverable! 
The exception here is a runtime crash, reloading the code to something that does not crash would then recover the App. 
The reason for the NSME is probably an issue with lambdas. We're actively working on the compiler and runtime to fix such issues. cc @AbdullinAM : Let’s add such a case to our tests and see if any of the current patches would already fix this.

natanfudge:
@sellmair But it's not recovering; The window closes! It should stay in the old state instead of closing if it plans to recover.

PR #180, opened 2025-05-13
Draft: #169, #170, #178: Don't allow to redefine compose entry method during reload

AbdullinAM:
### Core idea:
* Track user methods that are entry points to the Compose application
* When resolving dirty methods, if any of them is an entry point method &mdash; fail with an error

### Example:

```kotlin
fun main() {
    /**
     * Any change here is not allowed, it happens outside Compose scope,
     * and we can't reliably guarantee correctness
     * 
     * Any change here will mark `MainKt.main()V` as dirty 
     */
    application {
        /** 
         * Any change here is allowed, it happens inside Compose scope
         * 
         * Any change here will mark `MainKt$1.invoke()V` as dirty,
         * but will not affect `MainKt.main()V`
         */
        Window(::exitApplication) {
            Text("gg")
        }
    }
}
```


hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/RuntimeInfo.kt
```diff
@@ -117,21 +118,40 @@ internal fun ClassInfo(classNode: ClassNode): ClassInfo? {
...
 }
+
+private val COMPOSE_APP_ENTRIES = setOf(
+    Ids.WindowDesktopKt.singleWindowApplication,
+    Ids.WindowDesktopKt.singleWindowApplication_default,
+    Ids.ApplicationDesktopKt.application,
+    Ids.ApplicationDesktopKt.application_default,
+    Ids.ScreenshotTestApplicationKt.screenshotTestApplication,
+    Ids.ScreenshotTestApplicationKt.screenshotTestApplication_default,
+)
```
sellmair: Idea: I think we could hardcode those, that's fine. But maybe there is an elegant solution: 
Every time a 'Non Composable' function calls into a function, passing a Composable Lambda would be illegal, right? Just unclear to me how easy this would be?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/RuntimeInfo.kt
```diff
@@ -117,21 +118,40 @@ internal fun ClassInfo(classNode: ClassNode): ClassInfo? {
...
+
+private fun isComposeEntry(methodNode: MethodNode): Boolean {
+    return methodNode.instructions.any {
+        it is MethodInsnNode &&
+            MethodId(ClassId(it.owner), it.name, it.desc) in COMPOSE_APP_ENTRIES
```
sellmair: There is a factory function: `MethodId(it)`

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/RuntimeInfo.kt
```diff
@@ -117,21 +118,40 @@ internal fun ClassInfo(classNode: ClassNode): ClassInfo? {
...
-    return createRuntimeScopeInfo(methodId, runtimeInstructionTree)
+    val isComposeEntry = isComposeEntry(methodNode)
+    return createRuntimeScopeInfo(methodId, runtimeInstructionTree, isComposeEntry)
 }
 
 
 internal fun createRuntimeScopeInfo(
     methodId: MethodId,
     tree: RuntimeInstructionTree,
+    isComposeEntry: Boolean,
```
sellmair: Maybe we can actually carry more information, like: 
```
enum class MethodType {
    Regular, 
    ComposeEntryPoint,
    Composable, 
   
}
```

wdyt?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
```diff
@@ -49,7 +49,16 @@ private fun RuntimeInfo.resolveDirtyMethods(redefined: RuntimeInfo): List<Method
     return redefined.methodIndex.mapNotNull { (methodId, redefinedMethod) ->
         val previousMethod = methodIndex[methodId] ?: return@mapNotNull redefinedMethod
         if (previousMethod.rootScope.hash != redefinedMethod.rootScope.hash) {
-            return@mapNotNull redefinedMethod
+            if (redefinedMethod.rootScope.isComposeEntry) {
+                throw IllegalStateException(
+                    """
+                Compose Hot Reload does not support the redefinition of compose entry method $methodId.
```
sellmair: Some minor adjustments would be nice:

1) 'of the Compose entry method'
2) Please restart the App or revert the changes in '$methodId'

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ErrorRecoveryTests.kt
```diff
@@ -126,4 +126,68 @@ class ErrorRecoveryTests {
             fixture.checkScreenshot("3-after-change")
         }
     }
+
+    @HotReloadTest
+    @QuickTest
+    fun `illegal code change - update compose entry function`(fixture: HotReloadTestFixture) = fixture.runTest {
```
sellmair, line 132: I think the test is great, maybe we should think of this whole concept of 'rejecting reloads' as a concept like 'validation'? Then could have a separate validation test and one test here that tests that we can actually recover from a broken validation?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
```diff
@@ -49,7 +49,16 @@ private fun RuntimeInfo.resolveDirtyMethods(redefined: RuntimeInfo): List<Method
     return redefined.methodIndex.mapNotNull { (methodId, redefinedMethod) ->
         val previousMethod = methodIndex[methodId] ?: return@mapNotNull redefinedMethod
         if (previousMethod.rootScope.hash != redefinedMethod.rootScope.hash) {
-            return@mapNotNull redefinedMethod
+            if (redefinedMethod.rootScope.isComposeEntry) {
```
sellmair: I think the exception works pretty well here. 
What do you think about trying to formalize the concept as 'validation' (or any other name) and giving it its own space? If we keep the code here maybe we should return a `Try` instead of throwing? Or an `Either` that models the rejection cases.

sellmair (changes requested):

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
AbdullinAM: Yes, I think extracting the validation as a separate step during the reload is a very good idea. I will work on that

Issue #183, opened 2025-05-16
`NullPointerException` when attempting an invalid change in the Compose entry method


AbdullinAM:
### Reproduce:

1. Run Sample/Counter: Main
2. Replace
```
installEvas(events, states) {
    App()
}
```
With
```
installEvas(events, States()) {
    App()
}
```

Save the file

3. Revert back the changes, save the file

### Error
```
Error was captured in composition while live edit was enabled.
java.lang.NullPointerException: Parameter specified as non-null is null: method io.sellmair.evas.compose.CompositionLocalsKt.installEvas, parameter states
	at io.sellmair.evas.compose.CompositionLocalsKt.installEvas(CompositionLocals.kt)
	at MainKt$main$1$2.invoke(Main.kt:25)
	at MainKt$main$1$2.invoke(Main.kt:23)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:130)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
	at org.jetbrains.compose.reload.jvm.JvmDevelopmentEntryPoint$setContent$1$1.invoke(DevelopmentEntryPoint.kt:102)
	at org.jetbrains.compose.reload.jvm.JvmDevelopmentEntryPoint$setContent$1$1.invoke(DevelopmentEntryPoint.kt:101)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
```

AbdullinAM:
Smaller reproducer:
1. Run
```
fun main() {
    val str1 = ""
    val str2 = ""
    singleWindowApplication {
        Text(str1 + str2)
    }
}
```

2. Change to
```
fun main() {
    val str1 = ""
    val str2 = ""
    singleWindowApplication {
        Text(str1 + "str2")
    }
}
```
3. Return back to
```
fun main() {
    val str1 = ""
    val str2 = ""
    singleWindowApplication {
        Text(str1 + str2)
    }
}

**Comment by AbdullinAM** [comment]
Even smaller reproducer.
Before:
```
fun main() {
    val str1 = "aaaa"
    singleWindowApplication {
        Text(str1)
    }
}
```

After:
```
fun main() {
    val str2 = "aaaa"
    singleWindowApplication {
        Text(str2)
    }
}
```

AbdullinAM:
Fixed in #184

PR #184, opened 2025-05-19
#183: Don't allow changing local variables in compose entry method

AbdullinAM:


tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/VerificationTests.kt
```diff
@@ -84,11 +84,60 @@ class VerificationTests {
...
+    fun `illegal code change - change local variables in main`(fixture: HotReloadTestFixture) = fixture.runTest {
+        val code = fixture.initialSourceCode(
+            """
+                import androidx.compose.material.Text
+                import org.jetbrains.compose.reload.test.*
+                
+                fun main() {
+                    val str1 = "Hello"
+                    screenshotTestApplication {
+                        Text(str1 + "world!")
```
sellmair: Please make sure to use 'TestText' instead of 'Text'.
'Text' might render slightly differently on different platforms or machines. 
TestText tries to be more consistent

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/VerificationTests.kt
```diff
@@ -84,11 +84,60 @@ class VerificationTests {
...
+            """.trimIndent()
+        )
+
+        fixture.checkScreenshot("0-initial")
+
+        fixture.runTransaction {
+            /*
+            Legal Change: replace the code inside the composable function
+            */
+            code.replaceText(""""world!"""", """" world!"""")
```
sellmair: When changing code, make sure to use more 'different-looking' texts: 
The image diff algorithm is OK with anti-aliasing working different and therefore is lenient towards 'some difference'. It would be better for the human eye and the algorithm if we use something like 'Foo' & 'Bar' or 'A' 'B' or

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/RuntimeCodeHash.kt
```diff
@@ -102,11 +88,41 @@ internal fun RuntimeInstructionTree.codeHash(): RuntimeInstructionTreeCodeHash {
...
+        crc.pushHash(child.type.name)
+        crc.pushHash(child.group?.key)
+        crc.pushHash(child.failure?.message)
+        crc.pushHash(child.failure?.throwable?.stackTraceToString())
+    }
+
+    return RuntimeCodeHash(crc.value)
+}
+
+internal fun MethodNode.localVariablesHash(): RuntimeCodeHash {
```
sellmair: Why not just put that into the regular 'code hash' and call it a day?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/RuntimeInfo.kt
```diff
@@ -120,24 +120,24 @@ internal fun RuntimeScopeInfo(
...
+        methodType = MethodType(methodNode),
         scopeType = tree.type,
         group = tree.group,
         methodDependencies = tree.methodDependencies(),
         fieldDependencies = tree.fieldDependencies(),
-        children = tree.children.map { child -> createRuntimeScopeInfo(methodId, child, methodType) },
-        hash = tree.codeHash(),
+        children = tree.children.map { child -> createRuntimeScopeInfo(methodId, methodNode,child) },
+        codeHash = tree.codeHash(),
+        variableHash = methodNode.localVariablesHash(),
```
sellmair: If we really need to separate codeHash from 'variableHash', then it becomes a little hard to understand what is actually what and what should be done with it: We should document it then?
However, putting this information into the 'hash' would make it really simple imho

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/VerificationTests.kt
```diff
@@ -84,11 +84,60 @@ class VerificationTests {
...
+            requestReload()
+            val request = skipToMessage<OrchestrationMessage.ReloadClassesRequest>()
+            val result = skipToMessage<OrchestrationMessage.ReloadClassesResult>()
```
sellmair: In this case I think there is a 'awaitReload' maybe even a 'reload()' which is 'requestReload' and 'awaitReload'?

sellmair (approved):

sellmair (approved):

PR #185, opened 2025-05-20
ReloadClassesResult orchestration message stores the type of the result, including verification error

AbdullinAM:


hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
```diff
@@ -213,14 +213,25 @@ public sealed class OrchestrationMessage : Serializable {
...
         }
+
+        public enum class ResultType {
+            Success, VerificationError, Failure;
+
+            internal companion object {
+                @Suppress("unused")
+                internal const val serialVersionUID: Long = 0L
+            }
+        }
```
sellmair: So far, I think the only usage of this is in the devtools to render a different message, right? Are you sure you want to change (and commit) to this breaking change to the orchestration? 

Alternatively, the errorMessage carried here could just be groomed before sending the message!

sellmair (commented):
Generally looks good! Thanks for adding the test. However, please carefully evaluate if the change in the orchestration protocol is really desired.

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
AbdullinAM: Yes, this was also one of my concerns. Thats why I wanted to hear your opinion on it. I feel like adding more explicit result type can be (potentially) benefitial, but we don't have any real use cases for it right now

PR #203, opened 2025-05-26
Linter cleanup

qwwdfsad:
While looking at the code, I've decided to stretch IJ a bit and use its quickfixes on everything that is reported and that doesn't change the semantics.

The PR contains mostly mechanical fixes.
Testing: `./gradlew build` and nothing more (not sure if anything is expected)

### DIFF
--- a/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
+++ b/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
@@ -83,7 +83,7 @@ internal fun reload(
                     appendLine("⚠️ Superclass: '${originalClass.superclass?.name}' -> '${clazz.superclass?.name}'")
                 }
 
-                val addedInterfaces = clazz.interfaces.map { it.name } -
+                val addedInterfaces = clazz.interfaces.map { it.name }.toSet() -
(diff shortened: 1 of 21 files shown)


hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
```diff
@@ -84,7 +84,7 @@ internal fun reload(
                 }
 
                 val addedInterfaces = clazz.interfaces.map { it.name } -
```
AbdullinAM: I suggest converting `clazz.interfaces.map { it.name }` to set as well

hot-reload-test/src/main/kotlin/org/jetbrains/compose/reload/test/systemProperties.kt
```diff
@@ -12,7 +12,7 @@ internal val compileClasspath = requireEnvVar("chr.compilePath").split(File.path
 
 internal val compilePluginClasspath = requireEnvVar("chr.compilePluginPath").split(File.pathSeparator).map(::Path)
```
AbdullinAM, line 13: Maybe we can add explicit type to the other global variables in this file as well?

PR #206, opened 2025-05-27
Logging rework

AbdullinAM:
* Removed `slf4j` and `logback` dependencies, so that our logs do not interfere with user logs
* Created custom `HotReloadLogger` interface which is now used everywhere within the project
* All the logging is now done through orchestration: server receives all log messages and saves them to the log file
  * On the "user" side all the logging is done through slf4j-like API
* Stdout logging is disabled by default, but can be enabled with `compose.reload.enableStdoutLogging` option
  * Stdout enabled by default for tests
* All the log messages (including orchestration) now consider allowed logging level, can be configured with `compose.reload.loggingLevel` property (`Debug` by default)
* Created custom `HotReloadLoggerService` with one provided implementation

In addition:
* Unified way to access current orchestration handle through `OrchestrationService`


hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/devTools.kt
```diff
@@ -9,45 +9,29 @@ import org.jetbrains.compose.reload.core.HotReloadEnvironment
...
 
-    val process = ProcessBuilder(
+    ProcessBuilder(
         resolveDevtoolsJavaBinary(), "-cp", classpath.joinToString(File.pathSeparator),
         *subprocessDefaultArguments(DevTools, orchestration.port).toTypedArray(),
         *issueNewDebugSessionJvmArguments("DevTools"),
         "-Dapple.awt.UIElement=true",
         "org.jetbrains.compose.devtools.Main",
     ).withHotReloadEnvironmentVariables(DevTools).start()
-
```
sellmair, line 38: Am I right, that this was removed because of a potential duplication:
DevTools also using our logger will also send the messages into the orchestration. This would just send it again? In this case I would, however, I am afraid that we're missing parts (especially the stderr). Do you have an idea to capture the output still?

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/logging/CHRLogFormatter.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.jetbrains.compose.reload.logging
+
+import java.time.LocalDateTime
+
+
+public fun CHRLogFormatter(logger: HotReloadLogger): CHRLogFormatter = CHRDefaultFormatter(logger)
+
+public interface CHRLogFormatter {
```
sellmair: Personally, I think we could also get away with two static functions to format log messages. Are you planning to support different formats?

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/logging/CHRLogFormatter.kt
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright 2024-2025 JetBrains s.r.o. and Compose Hot Reload contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.compose.reload.logging
+
+import java.time.LocalDateTime
+
+
+public fun CHRLogFormatter(logger: HotReloadLogger): CHRLogFormatter = CHRDefaultFormatter(logger)
```
sellmair: Note: This applies to more places: 
I think we should neither use 'HotReload' nor 'CHR' as prefixes. For us it can just be the 'logger'. Wdyt? If we really see a reason for prefixing, then we could agree on one prefix, but so far I think we can get away by keeping it without classifier

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/logging/HotReloadLoggerService.kt
```diff
@@ -0,0 +1,30 @@
...
+ */
+@Suppress("NOTHING_TO_INLINE")
+public inline fun HotReloadLogger(): HotReloadLogger {
+    val clazz = MethodHandles.lookup().lookupClass()
+    return HotReloadLogger(clazz.name)
+}
+
+public fun HotReloadLogger(name: String, tag: String? = null): HotReloadLogger {
+    return ServiceLoader.load(HotReloadLoggerService::class.java)
+        ?.singleOrNull()
```
sellmair: Strongly think that we should use 'firstOrNull'. The first one listed wins: Would be a surprise if two registered services are available but we decide to not pick any of them? WDYT?

sellmair: Btw, it seems we have a composite logger in this MR: How about merging all loggers from all services?

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/logging/HotReloadLogger.kt
```diff
@@ -0,0 +1,221 @@
...
+
+public inline fun <reified T : HotReloadLogger> HotReloadLogger.with(): HotReloadLogger {
+    val other = T::class.java.getDeclaredConstructor(
+        String::class.java, Level::class.java
+    ).newInstance(name, level)
+    return this.with(other)
+}
+
+public fun HotReloadLogger.with(other: HotReloadLogger): HotReloadLogger = when {
+    this is HotReloadMultiLogger && other is HotReloadMultiLogger -> HotReloadMultiLogger(
```
sellmair: Love it 😍. What happened to the formatting here though?

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/logging/HotReloadLogger.kt
```diff
@@ -0,0 +1,221 @@
...
+        name,
+        level,
+        loggers + other.loggers
+    )
+    this is HotReloadMultiLogger -> HotReloadMultiLogger(name, level, loggers + other)
+    other is HotReloadMultiLogger -> HotReloadMultiLogger(name, level, listOf(this) + other.loggers)
+    else -> HotReloadMultiLogger(name, level, listOf(this, other))
+}
+
+public interface HotReloadLogger {
```
sellmair: The interface looks quite mighty. I assume it started by sticking to slf4j. Do you think it would be worth reducing the surface area a little?

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/orchestration.kt
```diff
@@ -5,28 +5,36 @@
...
+internal class AgentOrchestrationService : OrchestrationService {
+    override fun getOrchestration(): OrchestrationHandle = _orchestration
 
-val orchestration by lazy { startOrchestration() }
+    private companion object {
+        private val _orchestration by lazy { startOrchestration() }
+    }
+}
+
+internal val orchestration: OrchestrationHandle = OrchestrationHandle()
```
sellmair: Am I right, the idea here is to push us towards using the Service to get the 'OrchestrationHandle' instead of directly accessing the orchestration?

I see some issues here (and on the 'OrchestrationHandle' site): 
There are just two environments where we have a staticly bound 'OrchestrationHandle' available: All other environments will crash when trying to call `OrchestrationHandle()`. 

I think its safer to use the static orchestration instances directly, if available, and only use the service to provide them (optionally) if we're in such an environment.

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
```diff
@@ -76,7 +76,7 @@ internal fun reload(
                 return@mapNotNull null
             }
 
-            logger.orchestration(buildString {
+            logger.info(buildString {
```
sellmair: Idea: What do you think about having a `logger.info {}` which gets a string builder as receiver. The lambda will only be invoked when logging is even enabled for that level. Not sure if we should do this, because the logger is quite mighty already and I requested to reduce its scope

hot-reload-test/gradle-testFixtures/src/main/kotlin/org/jetbrains/compose/reload/test/gradle/orchestrationTestLogging.kt
```diff
@@ -44,13 +44,16 @@ or finished.
...
-    Job() + Dispatchers.IO + CoroutineName("Orchestration Test Logging") + CoroutineExceptionHandler { ctx, e ->
+    Job() + Dispatchers.IO + CoroutineName("Orchestration Test Logging") + CoroutineExceptionHandler { _, e ->
         e.printStackTrace()
     }
 )
 
 @OptIn(ExperimentalPathApi::class)
 internal fun ExtensionContext.startOrchestrationTestLogging(server: OrchestrationServer) = testLoggingScope.launch {
+    // enable logging to stdout for tests
+    System.setProperty("compose.reload.enableStdoutLogging", "true")
```
sellmair: Such system properties are better to be put into the build.gradle.kts when launching the tests, therefore its easier to reason about the env

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/logging/HotReloadLoggerService.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2024-2025 JetBrains s.r.o. and Compose Hot Reload contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.compose.reload.logging
```
sellmair: I think it would be better to keep it under .compose.reload.core.* as well
Either flat in .core or in .core.logging

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/HotReloadOrchestrationLoggerService.kt
```diff
@@ -0,0 +1,35 @@
...
+    override fun getLogger(name: String, tag: String?): HotReloadLogger {
+        val logger = createLogger(name)
+        val tag = when {
+            tag != null -> tag
+            name.startsWith("org.jetbrains.compose.reload.agent") -> TAG_AGENT
+            name.startsWith("org.jetbrains.compose.devtools") -> TAG_DEVTOOLS
+            name.startsWith("org.jetbrains.compose.reload.jvm") -> TAG_RUNTIME
+            else -> null
+        }
+        val orchestration = Try { OrchestrationHandle() }.getOrNull()
```
sellmair: This is a good line which supports my intuition that `OrchestrationHandle` (or any access through the service) shall be optional by nature!

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/ConsoleLogState.kt
```diff
@@ -18,15 +18,14 @@ data class ConsoleLogState(val logs: List<String>) : State {
...
-
 }
 
 internal fun CoroutineScope.launchConsoleLogState() = launchState(ConsoleLogState) {
     val logDeque = ArrayDeque<String>()
 
     orchestration.asFlow().collect { event ->
         if (event is OrchestrationMessage.LogMessage) {
-            logDeque.add("${event.tag} | ${event.message}")
+            logDeque.add("${event.tag} | ${event.message.substringAfterLast("] ")}")
```
sellmair: I do not think this is a good idea, my messages can easily contain this character. Feel free to change the Orchestration.LogMessage for this MR

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/main.kt
```diff
@@ -62,6 +60,8 @@ fun main() {
         if (HotReloadEnvironment.devToolsIsHeadless) return@application
         if (!HotReloadEnvironment.devToolsEnabled) return@application
 
+        val logger = HotReloadLogger()
```
sellmair: Innocent change: Am I right, that the logger needs to be created after the 'setupOrchestration' process or it will not receive the `OrchestrationHandle`?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/orchestration.kt
```diff
@@ -9,29 +9,30 @@ import androidx.compose.runtime.Composable
...
+    override fun getOrchestration(): OrchestrationHandle = _orchestration
 
-internal val orchestration: OrchestrationHandle = run {
-    val handle = ServiceLoader.load(OrchestrationExtension::class.java)
-        .firstNotNullOfOrNull { extension -> extension.getOrchestration() }
-        ?: (OrchestrationClient(OrchestrationClientRole.Tooling))
-        ?: run {
-            logger.error("Failed to create orchestration client")
-            shutdown()
+    private companion object {
```
sellmair: OUt of curiosity: Why the companion object here?

hot-reload-runtime-jvm/src/dev/kotlin/org/jetbrains/compose/reload/jvm/DevApplication.kt
```diff
@@ -28,7 +28,7 @@ internal fun main(args: Array<String>) {
     try {
         run(args)
     } catch (t: Throwable) {
-        orchestration.sendMessage(OrchestrationMessage.CriticalException(Application, t))
+        OrchestrationHandle().sendMessage(OrchestrationMessage.CriticalException(Application, t))
```
sellmair: I think this could lead to a leaking process, right?
The `OrchestrationHandle` could fail, the main thread dies here before the exitProcess and non-daemon threads will keep the process alive. 
Maybe another good indicator that this should be optional (or Try)

sellmair:
When running a test I see log messages like:
```text
09:18:19.05: Log [null] [2025-05-28T09:18:19.053208] [Info] RuntimeDirtyScopes resolved in [1.729209ms] 
                                ^                            ^
                          tag missing?             lets format time to only be hh:mm:ss.(ms)
                          
```

properties.yaml
```diff
@@ -264,3 +264,20 @@ LaunchMode:
   target: [ application, devtools ]
   documentation: |
     Tells the application 'how' it was launched
+
+EnableStdoutLogging:
```
sellmair, line 268: Both properties would be convenient if wired from the build into the application: 
See the 'withComposeHotReloadArguments' function to do this

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/logging/CHRLogFormatter.kt
```diff
@@ -0,0 +1,47 @@
...
+}
+
+internal class CHRDefaultFormatter(
+    private val logger: HotReloadLogger
+) : CHRLogFormatter {
+    private fun time(): LocalDateTime = LocalDateTime.now()
+
+    override fun format(level: Level, msg: String?, t: Throwable?): String {
+        return buildString {
+            append("[${time()}] ")
```
sellmair: Feature Request: Colors 😍

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/devTools.kt
AbdullinAM: Yes, the idea was to avoid duplication and just avoid having to manually redirect all the output using this method.

I don't think we need stdout right now, because nothing that we don't already have should be printed there. I will return the redirecting of stderr

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/orchestration.kt
AbdullinAM: I wanted to have a single `OrchestrationHandle` instance for the whole process. If we make `_orchestration` just a class property there, there will be multiple instances of `_orchestration`, mainly because of loggers

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/main.kt
AbdullinAM: In the original version, actually, it was still OK. Creating a Logger would initiate a call to `OrchestrationHandle`, which will load the orchestration service and create a connection.

I moved it here just to avoid unnecessary confusion

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
AbdullinAM: This can be implemented as extension functions, which will not directly increase the interface's size. I think it is a good idea

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/runtimeTracking.kt
```diff
@@ -73,14 +73,10 @@ internal fun enqueueRuntimeAnalysis(
         classLoaders[classInfo.classId] = WeakReference(loader)
 
         if (classBeingRedefined == null) {
-            if (logger.isTraceEnabled) {
```
sellmair, line 76: I think this might be called very often and maybe actually not creating the string mgiht be beneficial? (But just a gut feeling) e.g.l using `trace {}`

hot-reload-test/gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/test/functionalTestTask.kt
```diff
@@ -74,6 +74,11 @@ internal fun Project.configureGradleTestTasks() {
             task.screenshotsDirectory.map { it.asFile.absolutePath }.string()
         )
 
+        task.systemProperty(
+            HotReloadProperty.EnableStdoutLogging.key,
+            "true"
```
sellmair: I am a little bit confused why this was added, as it seems to have no effect?
The functionalTest tasks imho are not nice for stdout logging  anyways as they have this higher level of parallelism. I expect only the log message pointing to the log files to be printed, but now this message is missing?

sellmair:
Here is some feedback on the logs:



```
[Compiler]: [13:01:02.177] [java.lang.ProcessBuilder] [Debug] > Task :app:assembleJvmMainResources UP-TO-DATE
```
- java.lang.ProcessBuilder: Seems like this logger name is not nice!
- IIRC the log level (aka the `[Debug]` part is usually in front, is it?

sellmair:
Closing this in favor of a new branch inspired by the changes and discussions of this effort.

Issue #156, opened 2025-04-30
Can the DevTools be displayed on top of other windows, maybe via a setting?


Dima-369:
On macOS, when I launch the Compose application on the right side, and IntelliJ on the left (I am using Hammerspoon window split commands), IntelliJ overlaps the DevTools indicator, and I am unable to see the DevTools, at all, while staying inside IntelliJ.

It would be very useful to have, especially to see occurring errors.

sellmair:
Good Idea! I think we should do this

AbdullinAM:
The original idea of 'always on top' mode for devtools window was replaced by the idea of 'detached mode' for devtools window. In detached mode, dev tools runs just as a regular app with decorated window and can be manually moved to the desired place on the screen.

Implemented in #229

PR #207, opened 2025-05-27
#156 - Detached mode for devtools window

AbdullinAM:
~~Dev tools window always displayed on top by default. This behaviour can be disabled using configuration option. I'm was not sure about the necessity of having that option, but I decided that it can be useful in situations with multiple window apps. Experience of having several dev tool windows always floating around can be confusing~~

Setting `compose.reload.devToolsAlwaysOnTop=true` will enable dev windows always on top mode


properties.yaml
```diff
@@ -167,6 +167,15 @@ DevToolsTransparencyEnabled:
     Some platforms might not be able to render transparency correctly (e.g. some linux environments).
     This property will allow such platforms to disable/enable transparency
 
+DevToolsAlwaysOnTop:
+  key: compose.reload.devToolsAlwaysOnTop
+  type: boolean
+  default: "true"
```
Nohus: It should default to false in my opinion. It's confusing if the app window is behind another window, but you can see the tool window next to "nothing" regardless.

AbdullinAM: Yes, I agree. Thinking about the experience of users that want to see it always on top I forgot to consider the experience of users that don't expect that :)

sellmair:
If we would like to have this feature, then we should loop the property from the build into the running application! (See `withComposeHotRelaodArguments`

AbdullinAM:
@sellmair I think this option can be useful in situations like this:
Screenshot 2025-06-02 at 10 41 34

You still want to see the devtools icon, but you may not want to change the behaviour of the whole app

sellmair:
How about then having just one 'detached window' mode?

AbdullinAM:
That actually sounds really interesting!

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/arguments.kt
```diff
@@ -42,6 +43,7 @@ sealed interface ComposeHotReloadArgumentsBuilder {
     fun setDevToolsEnabled(enabled: Provider<Boolean>)
     fun setDevToolsClasspath(files: FileCollection)
     fun setDevToolsTransparencyEnabled(enabled: Provider<Boolean>)
+    fun setDevToolsDetached(alwaysOnTop: Provider<Boolean>)
```
sellmair: I think the 'alwaysOnTop' name is a leftover: 
Maybe `isDetached`?

sellmair:
What is the current state of this PR?

AbdullinAM:
@sellmair It was stale for a bit, because I waited for #213 to be merged. I started working on in again, will be ready for review in a couple of days

sellmair:
Checklist for landing these changes

- [ ] Window should have a tray icon and support minimizing
- [ ] Window has an integrated window decoration with default close, minimize, and maximize icons from the platform
- [ ] System Property is wired through from Gradle

sellmair (changes requested):
Checklist for landing these changes

- [ ] Window should have a tray icon and support minimizing
- [ ] Window has an integrated window decoration with default close, minimize, and maximize icons from the platform
- [ ] System Property is wired through from Gradle

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/devTools.kt
```diff
@@ -63,3 +67,20 @@ private fun resolveDevtoolsJavaBinary(): String? {
 
     return null
 }
+
+private fun platformSpecificJvmArguments(): Array<String> = when (Os.currentOrNull()) {
```
sellmair: We could use `Os.current()` instead and throw an exception. I think this would be fair. WDYT? Would you rather be defensive here and only log the issue?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtAnimatedWindow.kt
```diff
@@ -0,0 +1,182 @@
...
+import org.jetbrains.compose.reload.core.HotReloadEnvironment.devToolsTransparencyEnabled
+import org.jetbrains.compose.reload.core.WindowId
+import org.jetbrains.compose.reload.core.createLogger
+import org.jetbrains.compose.reload.core.debug
+import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.ApplicationWindowGainedFocus
+import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.ShutdownRequest
+import java.awt.Dimension
+import java.awt.Point
+import kotlin.system.exitProcess
+import kotlin.time.Duration.Companion.milliseconds
```
sellmair, line 35: The aesthetics of this source file is really appreciated by me 😍

Issue #200, opened 2025-05-26
DevTools: Fix Expand/Minimize animation flicker


sellmair:

AbdullinAM:
Current status:
* #213 fixed the window resize flicker
* Window expand/minimize flicker is a known issue in CMP Desktop. Its status can be tracked [here](https://youtrack.jetbrains.com/issue/CMP-7700/Desktop-Synchronize-Compose-and-Swing-layout-and-rendering-phases). The only workaround is to remove the window size animation. Suggestion by @sellmair: lets make two separate windows --- one for minimised state and one for expanded

PR #213, opened 2025-06-05
#200: Fix dev tools window resize flicker

AbdullinAM:
Fix the animation flicker using a [workaround](https://youtrack.jetbrains.com/issue/CMP-8305) suggested by Compose Multiplatform team: controlling window size and position directly, instead of using `DialogState`. 

Using `DialogState` requires recomposition before the size/position is changed, which causes flicker when the changes are very frequent.


sellmair:
Unfortunately, I saw this when trying to open the dev tools on my machine:

image

sellmair:
image

sellmair:
https://github.com/user-attachments/assets/f9049d8c-8be9-4459-a409-ac4a36f8d371

Can you try to reproduce this ?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtSidecarWindow.kt
```diff
@@ -49,34 +56,48 @@ import org.jetbrains.compose.reload.core.WindowId
...
+
 @Composable
 fun DtSidecarWindow(
     windowId: WindowId,
     windowState: WindowState,
     isAlwaysOnTop: Boolean,
 ) {
-
     var isExpanded by remember { mutableStateOf(false) }
+    val initialized = remember { mutableStateOf(true) }
```
sellmair: I noticed that I was confused about this variable: 
Shouldn't this be called 'isInitializing'? 

Tip: maybe using `by remember` would look nicer?

sellmair (changes requested):

AbdullinAM:
@sellmair Yeah, I reproduced them as well. I though I fixed them, but apparently not. I have an idea of whats happening, so it should not be too hard to fix :)

sellmair:
I still see initial flicker on initial startup, but the window positioning issue seems fixed

AbdullinAM:
@sellmair Current status of this PR:

1. Window resize flickering is fixed.
2. Window expand/minimise flickering is a known issue in CMP on desktop. It can be tracked with [this ticket](https://youtrack.jetbrains.com/issue/CMP-7700/Desktop-Synchronize-Compose-and-Swing-layout-and-rendering-phases). The only advise I got on how to deal with it &mdash; remove window size animation, which I think we can't really do. It will only make UX worse because we will have a big transparent window with no clear borders.

My suggestion:
* Finish the review of this PR and merge it so that the resize flicker fix lands on master
* Write a report in #200 detailing the current status and come back to this issue when there are updates on CMP side

WDYT?

sellmair:
I am fine with the changes. We can squash commits and get it into staging. 
However, I would have hoped for a solution for the flicker. 
How about a solution where the floaty and sidebar are one window and the expanded is another window which then shows and starts animating? With such an approach now window would need to be resized

PR #214, opened 2025-06-10
Fix reloading on compose 1.9.0+: correctly search for FunctionKeyMeta…

AbdullinAM:
After our change to `FunctionKeyMeta` annotation, in newer compose versions it will appear in `MethodNode.invisibleAnnotations`, instead of `MethodNode.visibleAnnotations`. So we can just check for both of them.

Alternatively, we can check for the current CMP version and decide which list to check, but I don't think it makes sense currently. @sellmair WDYT?

### DIFF
--- a/.teamcity/testDimensions.json
+++ b/.teamcity/testDimensions.json
@@ -23,8 +23,11 @@
   ],
   "compose": [
     {
-      "version": "1.8.0",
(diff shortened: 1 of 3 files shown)


sellmair:
Let's also bump the tests to run against the affected CMP version.

PR #221, opened 2025-06-12
Support replacing image resources

pjBooms:


tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ResourcesTests.kt
```diff
@@ -31,45 +32,61 @@ import kotlin.io.path.name
 @ExtendBuildGradleKts(ResourcesTests.Extension::class)
 class ResourcesTests {
```
sellmair, line 33: Thanks for adding those tests 😍
Do you think there will be issues when people switch from regular strings to plurals?

sellmair:
https://github.com/user-attachments/assets/b261207b-69a5-4e1f-b88c-6f9da48623ba

It seems that critical caches are not cleared yet. See this demo, where changing the size of the string either breaks the application or leads to the string value being partially decoded.

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
```diff
@@ -109,10 +113,33 @@ internal fun reload(
...
+private fun cleanResourceCache(classId: ClassId, cleanMethod: MethodId, resourceCacheType: String) {
+    val loader = findClassLoader(classId).get()
+    if (loader == null) {
+        logger.info("$resourceCacheType resource cache cleaning skipped: '$classId' is not loaded yet.")
+        return
+    }
+    loader.loadClass(classId.toFqn())
+        .getDeclaredMethod(cleanMethod.methodName)
+        .invoke(null)
+    logger.info("$resourceCacheType resource cache cleared")
```
sellmair: I would propose to wrap this in. `Try {}`:

A failure in resource invalidation shall be reported aggressively, but not break the reload. 
E.g. when compose resources breaks this integration in later changes, then we should not fail the reload, but report that the integration to compose resources is not working.

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
```diff
@@ -109,10 +113,33 @@ internal fun reload(
...
     instrumentation.redefineClasses(*definitions.toTypedArray())
-    return redefineRuntimeInfo().get().mapLeft { redefinition ->
+    return redefineRuntimeInfo(changedResources).get().mapLeft { redefinition ->
         val reload = Reload(reloadRequestId, definitions, redefinition)
         reinitializeStaticsIfNecessary(reload)
         reload
     }
 }
+
+private fun cleanResourceCache(classId: ClassId, cleanMethod: MethodId, resourceCacheType: String) {
```
sellmair: Similarly to handling statics, this code should likely live in its own source file. 

I can totally see that this can be an extension working in 'invokeAfterHotReload'

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/runtimeTracking.kt
```diff
@@ -41,11 +42,11 @@ internal fun launchRuntimeTracking(instrumentation: Instrumentation) {
     instrumentation.addTransformer(RuntimeTrackingTransformer)
 }
 
-internal fun redefineRuntimeInfo(): Future<Try<RuntimeDirtyScopes>> = runtimeAnalysisThread.submitSafe {
+internal fun redefineRuntimeInfo(changedResources: List<File> = emptyList()): Future<Try<RuntimeDirtyScopes>> = runtimeAnalysisThread.submitSafe {
```
sellmair: A change in dirty resolution is kind of expected, but I would not expect it at this stage 🤔 

The runtimeTracker should know about the current state and not require additional external information passed as a parameter. Can you think of a solution that won't require this parameter?

sellmair (changes requested):

pjBooms:
> Screen.Recording.2025-06-17.at.11.13.06.mov 
> It seems that critical caches are not cleared yet. See this demo, where changing the size of the string either breaks the application or leads to the string value being partially decoded.

The root cause of the problem is here: https://github.com/JetBrains/compose-hot-reload/issues/230

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ResourcesTests.kt
pjBooms: Do you suggest to add a test? I don't expect problems here.

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/runtimeTracking.kt
pjBooms: I thought about this.

While the information about changed classes comes from the `Instrumentation` and `reload` fills in the `instrumentation`, changes in resources are only known at `reload` and that's it. 

We may artificially create a concept of say `ChangedResourses` that `reload` will fill in with changed resources and then `runtimeTracker` will add a listener to `ChangedResources` and  listen about changes in resources and fill its own private field  of changed resources that in turn be accessed in ` redefineRuntimeInfo`… But it seems to be over-engineering.

I also thought about adding a field named "resources" to RuntimeInfo, but again it seems to be an artificial thing for now and who will fill that field, why do we need it? 

When we map resource accessors to its resources, to invalidate only required composables we will definetly need to re-work this stuff, but for now it seems enough to just know if resources are changed in general and just pass the information directly.  

If you have some other ideas, let's discuss.

sellmair:
Can you quickly point me to the static state that is stored in compose resources, which would need to be re-initialized?

pjBooms:
> Can you quickly point me to the static state that is stored in compose resources, which would need to be re-initialized?

Generated resource accessors like `generated.resources.CommonMainString0`

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
pjBooms: Wrapped with try {} catch(t: Throwable)

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/reload.kt
pjBooms: Moved to its own source file. Please check if it is what you mean.

sellmair (approved):
💪

PR #228, opened 2025-06-16
OrchestrationProtocol: Propose stable and evolvable protocol

sellmair:


hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/Task.kt
```diff
@@ -0,0 +1,246 @@
...
+        name: String? = null, context: CoroutineContext = EmptyCoroutineContext, body: suspend Task<T>.() -> T
+    ): Task<T>
+
+    override val key: CoroutineContext.Key<*> get() = Key
+
+    public companion object Key : CoroutineContext.Key<Task<*>>
+}
+
+
+public fun Task<*>.isActive(): Boolean = !value.isCompleted() && !isStopped()
```
AbdullinAM: We discussed during the demo that maybe this should be suspendable. You decided against it?

hot-reload-core/src/test/kotlin/FutureLincheck.kt
```diff
@@ -0,0 +1,63 @@
...
+    @Operation
+    suspend fun awaitWithContinuation(): Int {
+        return suspendCoroutine<Int> {
+            future.awaitWith(it)
+        }
+    }
+
+    @EnabledIfSystemProperty(named = "lincheck", matches = "true")
+    @Test
+    fun modelTest() {
```
AbdullinAM, line 51: This test fails for me currently. Is this expected behaviour?

AbdullinAM (commented):
Really clean implementation🔥! I only have a couple of small comments left after the demo.

hot-reload-core/src/test/kotlin/FutureLincheck.kt
sellmair: No, I reported this to the lincheck maintainer as its not clear if this is a bug in lincheck!

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/Task.kt
sellmair: Good point! I still think we should go with the safer option of yielding by default. 
Implemented!

PR #233, opened 2025-06-18
core: Implement logging infrastructure

sellmair:
Example Output: App with active logging
image

Example Output: Error
image

Log File:
image


AbdullinAM (approved):
Really great implementation! With the new orchestration it is super clean!

PR #234, opened 2025-06-18
Add Amper server support

singleton11:
In order to improve performance, we run Amper server, which basically is the builing service. It receives HTTP requests to rebuild the project and sends traces back using SSE. So it's some kind of pseudo-continious mode that allows us to save on hot spot compiler, classloading, caches warmup etc


sellmair:
Happy to see this PR 😍 
Some 'raw thought': The entire 'recompiler process being managed by hot reload' concept was implemented because we initially did not have IDE tooling and therefore relied on file changes to trigger a recompile. 

At some point, the project will move towards 'explicit builds', which means that the IDE triggers the reload when the user presses CMD+S within the IDE. I wonder if Amper really needs to run this extra loop as well, or if Amper could just do all of this controlling logic within Amper directly?

Amper is launching the APP, Amper knows how to request reloads, the IDE plugin knows when to trigger this reload,. 
WDYT about it conceptually? Let's meet as soon as you would like to continue working on it!

sellmair:
Closed in favor of #241

PR #240, opened 2025-06-23
#200 use two separate windows for minimised and expanded devtools window

AbdullinAM:
Implementation based on @sellmair's idea to use two separate windows for devtools


sellmair:
During my initial testing, I saw the close and expand buttons behave weirdly. The buttons do not react at certain times, and at other times, the click gets executed way after I click. 

While some stutter and flicker seems to be reduced, I still see flickering, now especially at startup!

https://github.com/user-attachments/assets/67e0a7c3-ab27-4389-8705-a8c99acfc781

Issue #242, opened 2025-06-24
WindowTransformation: Support 1.9.0-alpha03 'SwingWindow' and 'SwingDialog'


sellmair:
Breaking change: https://github.com/JetBrains/compose-multiplatform-core/pull/2141

PR #243, opened 2025-06-24
#242: Support 1.9.0-alpha03 'SwingWindow' and 'SwingDialog'

AbdullinAM:
* Support new 'SwingWindow' and 'SwingDialog' from CMP 1.9.0-alpha03
* Support dialog windows for previous versions of CMP


sellmair:
Good Job 👍 
Looks good! I think we still do not run tests on CI, which would detect the issue, right?
The tests only run locally because they are headless? 

We can think of a testing strategy later!

AbdullinAM:
> I think we still do not run tests on CI, which would detect the issue, right? The tests only run locally because they are headless?

@sellmair Yes, the tests are unfortunately still not headless, so they run only locally


> We can think of a testing strategy later!

Agreed 👍 

I will wait for CI and merge

sellmair:
Green 🟢 💚 ; Lets go 🚀

PR #247, opened 2025-06-25
Sebastian Sellmair Fix Missing 'ReloadState.Reloading' after switching to RecompilerExtension design

sellmair:


AbdullinAM:
@sellmair I have encountered a weird issue when recompiling: it seems like it tries to recompile the project multiple times after the changes are introduced. It results in a flicker in status bar/dev tools window. Does this reproduce for you?

https://github.com/user-attachments/assets/c16db6e7-2155-43b8-8132-718d73322e07

sellmair:
@AbdullinAM 
This might indicate an even worse issue: Can you check how many active Gradle Daemons there are?
I suspect that the recompiler API changes have also brought in some isuses with daemon cleanup!

PR #250, opened 2025-06-26
Sebastian Sellmair Implement 'compose.reload.resources' feature flag

sellmair:


hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/resources.kt
```diff
@@ -55,14 +57,15 @@ private fun cleanResourceCache(classId: ClassId, cleanMethod: MethodId, resource
...
 internal fun Context.cleanResourceCacheIfNecessary() {
-    if (hasChangedResources()) {
-        cleanResourceCache(Ids.ImageResourcesKt.classId, Ids.ImageResourcesKt.dropImageCache, "Images")
-        cleanResourceCache(
-            Ids.StringResourcesUtilsKt.classId,
-            Ids.StringResourcesUtilsKt.dropStringItemsCache,
-            "Strings"
-        )
-    }
+    if (!HotReloadEnvironment.composeReloadResourcesEnabled) return
```
pjBooms: Cleaning resource caches is quite safe, I don't think we need to put it under feature flag.

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/arguments.kt
```diff
@@ -213,6 +215,7 @@ private class ComposeHotReloadArgumentsImpl(
     /* Non API elements */
     private val virtualMethodResolveEnabled: Boolean,
     private val dirtyResolveDepthLimit: Int,
+    private val composeResourcesEnabled: Boolean,
```
pjBooms: rename to composeReloadResourcesEnabled?

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/arguments.kt
```diff
@@ -183,6 +184,7 @@ private class ComposeHotReloadArgumentsBuilderImpl(
             /* Non API elements */
             virtualMethodResolveEnabled = project.composeReloadVirtualMethodResolveEnabled,
             dirtyResolveDepthLimit = project.composeReloadDirtyResolveDepthLimit,
+            composeResourcesEnabled = project.composeReloadComposeReloadResourcesEnabled,
```
pjBooms: composeReloadComposeReloadResourcesEnabled looks strange (repeated). Can it be changed?

pjBooms (changes requested):

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/arguments.kt
sellmair: OK

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/arguments.kt
sellmair: This is a unfortunate clash
We use `composeReload` as a namespace for the project (e.g., `compose.reload` packages) or here as a prefix of a property. However, the name of the property is also ComposeReload. 

Either way, it will look stupid and I do not think we should spend too much energy here 😬

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/resources.kt
```diff
@@ -55,14 +60,16 @@ private fun cleanResourceCache(classId: ClassId, cleanMethod: MethodId, resource
...
+
+    Try {
         cleanResourceCache(Ids.ImageResourcesKt.classId, Ids.ImageResourcesKt.dropImageCache, "Images")
         cleanResourceCache(
             Ids.StringResourcesUtilsKt.classId,
             Ids.StringResourcesUtilsKt.dropStringItemsCache,
             "Strings"
         )
-    }
+    }.leftOr { error -> logger.error("cleanResourceCacheIfNecessary failed", error.exception) }
```
pjBooms, line 72: I am not sure if it is necessary to wrap to `Try` here as I wrapped `cleanResourceCache` with traditional `try`

pjBooms (approved):

PR #256, opened 2025-06-30
Update version number in README.md

uramnoil:
1.0.0-beta01
https://mvnrepository.com/artifact/org.jetbrains.compose.hot-reload/gradle-plugin


sellmair:
beta01 is not going to be advertised for now. We're preparing beta02 this week. Sorry the confusion. We're trying to roll out some breaking changes and therefore wanted to publish to versions in quick succession.

PR #265, opened 2025-07-02
Process Destroy: Await descendants

sellmair:


AbdullinAM (approved):
This approach seems good to me 👍  Should not cause any process leaks, unless something goes completely wrong

PR #271, opened 2025-07-03
Support replacing font resources

pjBooms:
Support replacing font resources in CHR that becomes possible after [the change](https://github.com/JetBrains/compose-multiplatform/pull/5341) in Compose.


hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/resources.kt
```diff
@@ -56,10 +57,47 @@ private fun cleanResourceCache(classId: ClassId, cleanMethod: MethodId, resource
     }
 }
 
+private fun cleanResourceCaches(): Boolean {
```
sellmair: What do you think about calling the method `tryCleanResourceCache` and return a `Try` instead?

sellmair (approved):

Issue #232, opened 2025-06-18
Dev tooling windows should support shortcuts for "Select All"


zsmb13:
Cmd+A or Ctrl+A should select all the output if the focus is in the text area

![Image](https://github.com/user-attachments/assets/d82c7550-c2ef-4e51-8ce0-d58776ebcb25)

AbdullinAM:
Implemented in #273

PR #273, opened 2025-07-03
#232 Button to copy all text from the dev tools windows

AbdullinAM:
Request from #232: add "select all" functionality to dev tools windows

Problem: Compose does not provide a built in way to support "select all". Any workaround with selection manager may interfere with user selection (e.g. create confusions when user tries to manually select something)

Workaround: Add 'Copy all' button to all windows where it might be necessary


zsmb13 (approved):
Works nicely, thank you!

Maybe it could be styled a tiny bit like the other buttons are, with some padding on the clickable area + rounded corners?

AbdullinAM:
@zsmb13 My original idea was to make them less 'noticeable' than the main 'functional' buttons of the dev tools. If you think that consistent style is better I will try it and share the results

sellmair (approved):

Issue #274, opened 2025-07-03
Indy: Parent scope gets invalidated on change


sellmair:
When using invokedynamic (e.g. using `-Xindy-allow-annotated-lambdas`), then we experienced that, under certain conditions, the parent scope gets wrongfully invalidated:

```kotlin
@Composable
fun Foo() {
    Card {
              var state = // ..
            Card { 
                   // use state
                   // <change>
            } 
    }
}
```

Changing at `<change>` causes the outer state to reset.

PR #275, opened 2025-07-03
ResolveDirty: Ensure @Composable functions act as 'dirty barrier' 

sellmair:
Related Issue:
https://github.com/JetBrains/compose-hot-reload/issues/274

- Enabled tests against more 'invokedynamic'
- Discovered one test failure which indicates that under certain conditions a dirty 'Composable' function will mark depending 'Composable' functions as dirty as well (this shall not happen)
- Also added a test fixture for working directly on the dirty resolution (should make it easy to work on that system)


AbdullinAM:
Nice solution for the problem and a really great explanation of it in the comments 💪 🔥

PR #286, opened 2025-07-09
Investigate the memory consumption of CI agents

AbdullinAM:


AbdullinAM:
* Set custom Gradle daemon idle timeout of 10 second in `GradleRunner`
* Fix small resource leaks in `checkScreenshot`
* Reduce number of unnecessary list allocations in `checkScreenshot`

Issue #300, opened 2025-07-25
Version 1.0.0-beta04 introduces issues with DevTools on Linux


Nohus:
There are two issues here:
- The minimized CHR button has a white section at the bottom
- The DevTools window has a big white background as the animation is playing out

These are new issues that are not present in `beta-03`. Both are caused by transparency being disabled on Linux, and were previously mitigated by disabling animations on Linux.

https://github.com/user-attachments/assets/0bdc6f9c-a206-47be-bdae-5492fd94cd2d

AbdullinAM:
@Nohus Hi! Thank you for reaching out! These changes are somewhat intentional: beta-04 includes some changes to window management of DevTools, mainly aiming to fix various animation/flicker issues caused by window resizing.

* Minimised devtools window was extended to correctly display reload counter. It is hidden when there are no reloads, thats why initially there is just a white section.

<img width="710" height="310" alt="Image" src="https://github.com/user-attachments/assets/fcf1503b-1682-4781-acda-bd6d1e39616a" />

* White background is also added to ensure there is no flicker during the expand/minimise animation.


We are still working on polishing the looks of DevTools window, so I will try to come up with a better solution for Linux.

AbdullinAM:
Addressed mentioned issues in #301:

* Reload counter is now visible by default when transparency is disabled, so there is no white section on the minimised devtools window
* Added `compose.reload.devToolsAnimationsEnabled` option that controls the window animations. When disabled devtools window will immediately switch between expanede/minimised state and the will be no big white background during transition

PR #301, opened 2025-07-28
#300: Improve DevTools window when transparency is disabled

AbdullinAM:
* Don't use rounded corners so that there is no white background visible
* Enable reload counter by default (so that there is no white section)
* Option `compose.reload.devToolsAnimationsEnabled` to enable/disable animations so that there is no big white background during transition. Option is enabled by default, maybe we can disable it for Linux?
* Increase minimised window size on windows so that the reload counter fully fits

Before:

https://github.com/user-attachments/assets/a068c307-b6f2-486e-b915-76b9b8cc2fed

After:

https://github.com/user-attachments/assets/b3d660bd-0753-4b4e-b649-fed83b154f7d

### DIFF
--- a/properties.yaml
+++ b/properties.yaml
@@ -175,6 +175,15 @@ DevToolsDetached:
   documentation: |
     If enabled, dev tools window will be detached from the main application
 
+DevToolsAnimationsEnabled:
+  key: compose.reload.devToolsAnimationsEnabled
+  type: boolean
+  default: "true"
+  target: [ application, build, devtools ]
+  documentation: |
+    If disabled, dev tools will not use window animations when switching between expanded/minimised states. 
+    Note: window contents will still be animated.
+
 IntelliJDebuggerDispatchPort:
   key: compose.reload.idea.debugger.dispatch.port
   type: int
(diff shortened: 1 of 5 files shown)


properties.yaml
```diff
@@ -175,6 +175,15 @@ DevToolsDetached:
   documentation: |
     If enabled, dev tools window will be detached from the main application
 
+DevToolsAnimationsEnabled:
+  key: compose.reload.devToolsAnimationsEnabled
```
sellmair, line 179: What are the cases where people would just want to use this property for their project?

sellmair (approved):

properties.yaml
AbdullinAM: I think mainly this property can be useful for cases when transparency is disabled (e.g. like in #300) If people do not want to wait for animations and wait for the white background --- they can choose to turn off the animations

sellmair: Okay for me! Maybe we should go over all properties again to check what is needed. There are quite a few now. But lets move forward here ☺️

PR #304, opened 2025-07-28
Add property to enable dev tools debugging

AbdullinAM:
`compose.reload.idea.debugger.dispatch.port.propagate`

I'm open to suggestions for a better name 😃. I was not able to come up with anything better


sellmair:
Naming Discussion engaged: 
`compose.reload.subprocessDebuggingEnabled`

PR #312, opened 2025-08-06
(Restart|Replace)GroupTokenizer: Allow LineNumber- and Label Nodes between key and method invocation

### LINKED ISSUE #311: StartReplaceGroup: Allow LineNumberNodes between LDC and InvokeVirtual
The following code was reported to not parse the start replace group correctly

```
"""
            @Composable
            fun targetEnterExit(param: String): String =
                key(param) {
                    if (param.isEmpty()) {
                        "Empty"
                    } else {
                        val state = remember { mutableStateOf("") }
                        state.value
                    }
                }
        """,
```

sellmair:
Fixes #311: 

Previously, the tokenizer expected an LDC (or another constant instruction) immediately adjacent to the method invocation. 
However, some code might lead to emitting a Label or Line Number between those instructions.

Thanks to @ShikaSD for reporting this.


hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/InstructionToken.kt
```diff
@@ -455,3 +453,14 @@ internal fun tokenizeInstructions(
...
     return tokens.toLeft()
 }
+
+
+private fun TokenizerContext.Consumer.findMethodInsn(): MethodInsnNode? {
+    return find { node ->
+        when (node) {
+            is MethodInsnNode -> node
+            is LabelNode, is LineNumberNode -> null
+            else -> return null
```
AbdullinAM, line 459: Unnecessary `return` here

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/InstructionToken.kt
```diff
@@ -455,3 +453,14 @@ internal fun tokenizeInstructions(
 
     return tokens.toLeft()
 }
+
+
+private fun TokenizerContext.Consumer.findMethodInsn(): MethodInsnNode? {
+    return find { node ->
+        when (node) {
+            is MethodInsnNode -> node
+            is LabelNode, is LineNumberNode -> null
```
AbdullinAM, line 458: What is the motivation for explicitly checking `LabelNode` and `LineNumberNode` here? If you want to keep it, maybe we can add a comment to explain it?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/InstructionToken.kt
```diff
@@ -283,18 +287,12 @@ private object EndRestartGroupTokenizer : InstructionTokenizer() {
...
-
-        if (expectedMethodInsn is MethodInsnNode &&
-            MethodId(expectedMethodInsn) == Ids.Composer.startReplaceGroup
-        ) {
-            val keyValue = expectedLdc.intValueOrNull() ?: return Failure(
-                "Failed parsing startReplaceGroup token: expected key value"
-            ).toRight()
-
-            return StartReplaceGroup(ComposeGroupKey(keyValue), listOf(expectedLdc, expectedMethodInsn)).toLeft()
+        val consumer = context.consumer()
```
AbdullinAM, line 289: `SourceInformationTokenizer::nextToken` also expects LDC + MethodInsn, maybe we can update it to use `findMethodInsn` as well?

AbdullinAM: Same with `EndToMarkerTokenizer`, it expects ILOAD + MethodInsn

AbdullinAM (approved):
Nice job 💪! I only have couple of small comments on future-proofing  other tokenizers that (currently) expect two instructions back-to-back

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/InstructionToken.kt
sellmair: The return null will return the method, whereas the `-> null` will continue searching!

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/InstructionToken.kt
sellmair: Those two node types can effectively be 'ignored', so we scan until we find either a `MethodInsnNode` , consuming the 'noise'

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/InstructionToken.kt
AbdullinAM: Ah, missed it. Return semantics for inline functions is very cool, but sometimes a bit confusing 😅

sellmair:
But I found another issue with this: 
If the LabelNode gets consumed into the 'Start{YX}' token, then it cannot be used as target for jumps later. 
It should really not be used for a jump, but its still super confusing to me.

AbdullinAM:
> But I found another issue with this:
> If the LabelNode gets consumed into the 'Start{YX}' token, then it cannot be used as target for jumps later.
> It should really not be used for a jump, but its still super confusing to me.

Do you have example of the code that produces this?

sellmair:
Yes, the code from the test will consume the Label L8 into the `StartReplaceGroup `
```
StartReplaceGroup(key=ComposeGroupKey(key=-1406269919))
            LdcInsnNode(-1406269919) [18]
            LabelNode(L8)
            LineNumberNode [-1]
            MethodInsnNode(androidx/compose/runtime/Composer.startReplaceGroup (I)V) [185]
```

However, I think its somewhat OK, because such a label node cannot be used as jump target (if used it will break)

AbdullinAM:
I think so too, this kinds of tokens should not be jump targets.

Maybe we can add additional verification step to `tokenizeInstructions` to check that all jumps have valid corresponding labels after tokenisation? Just so that if something goes wrong --- we will know about it?

AbdullinAM (approved):

PR #314, opened 2025-08-06
checkScreenshot: Implement 'Stas Gradient Compare' algorithm

sellmair:
The previous algorithm handled Gradients by using a hand-tuned penalty score. However, the global 'threshold' still might be hard to understand and could lead to strange edge cases.

The new algorithm is simpler:
Within a given area (radius), we're looking for reasonable pixel values nearby:

If we find a pixel equal or lower and one equal or higher in the area, then we can assume the pixel value to be within a gradient.

Thanks to 'stanislav.erokhin' for the algorithm.


hot-reload-test/gradle-testFixtures/src/main/kotlin/org/jetbrains/compose/reload/test/gradle/skikoUtils.kt
```diff
@@ -0,0 +1,194 @@
...
+        for(int dx = -radius; dx <= radius; dx++) {
+            for(int dy  = -radius; dy <= radius; dy++) {
+                float2 targetCoord = coord + half2(dx, dy);
+                if(targetCoord.x <= 0 || targetCoord.x >= size.x || targetCoord.y <= 0 || targetCoord.y >= size.y) {
+                    continue;
+                }
+                
+                half4 expectColor = expect.eval(targetCoord);
+                
+                lowPixelFoundR = lowPixelFoundB || expectColor.r < upperBoundB;
```
AbdullinAM: Probably a typo here, should be `upperBoundR` and `lowerBoundR`?

AbdullinAM: And `lowPixelFoundR` / `highPixelFoundR`

AbdullinAM (approved):
Very cool shader magic 🪄 

Found only a small typo in shaders, otherwise all good! We can wait for tests to get green and send it 👍

hot-reload-test/gradle-testFixtures/src/main/kotlin/org/jetbrains/compose/reload/test/gradle/skikoUtils.kt
sellmair: Fixed! Really good catch, thank you!

sellmair:
Unfortunately, a bunch of things I was proud of were removed because the algorithm suggested by Stas is just so much simpler and works really well for us!

hot-reload-test/gradle-testFixtures/src/main/kotlin/org/jetbrains/compose/reload/test/gradle/skikoUtils.kt
```diff
@@ -0,0 +1,194 @@
...
+                highPixelFoundR = highPixelFoundR || expectColor.r > lowerBoundR;
+                
+                lowPixelFoundG = lowPixelFoundG || expectColor.g < upperBoundG;
+                highPixelFoundG = highPixelFoundG || expectColor.g > lowerBoundG;
+                
+                lowPixelFoundB = lowPixelFoundB || expectColor.b < upperBoundB;
+                highPixelFoundB = highPixelFoundB || expectColor.b > lowerBoundB;
+                
+                if(lowPixelFoundR && highPixelFoundR && lowPixelFoundG && highPixelFoundG && lowPixelFoundB && highPixelFoundB){
+                   return half4(0.0, 0.0, 0.0, 1.0);
```
harrytmthy: NIT: Should we include alpha in the comparison, or clarify in the KDoc that it's not included? 🤔

sellmair: 🤔 Fair! If we include alpha, would you expect it to work exactly the same as any other channel?

sellmair: Updated; Handles alpha now as well!

PR #316, opened 2025-08-07
Orchestration: Allow DevTools to host the orchestration server

### LINKED ISSUE #306: TimeoutException kills CHR when making changes during debugger pause
If I pause on a breakpoint, and then make a change, CHR will try to reload, but it stalls, and eventually gives up with a `TimeoutException`. 

```
> Task :hotSnapshotMain
Info  | Gradle | StatusService: Failed to connect to '55822' for ':
Info  | Gradle | java.util.concurrent.TimeoutException
Info  | Gradle | 	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1960)
Info  | Gradle | 	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2095)
Info  | Gradle | 	at org.jetbrains.compose.reload.core.FutureKt.getBlocking-HG0u8IE(Future.kt:202)
Info  | Gradle | 	at org.jetbrains.compose.reload.orchestration.OrchestrationHandleKt.connectBlocking(OrchestrationHandle.kt:51)
Info  | Gradle | 	at org.jetbrains.compose.reload.gradle.StatusService.<init>(statusService.kt:92)
Info  | Gradle | 	at org.jetbrains.compose.reload.gradle.StatusService$Inject.<init>(Unknown Source)

```

After I resume the app, CHR is "dead" and I am forced to restart the app. 

I would expect CHR to either run during a breakpoint pause, not timeout from such pauses, or timeout and then resume operations normally after the pause is done (reloading when it can)

CHR 1.0.0-beta04
Gradle 8.14
Kotlin 2.2.0
Compose 1.8.0
Windows 11

sellmair:
Having the orchestration hosted by the user application
might lead to issues in situations where the user application
might be dead (or stopped by debugger).

Having the orchestration server hosted by devtools
will ensure smooth operations even in such situations.

Fixes #306

### DIFF
--- a/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/compose.kt
+++ b/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/compose.kt
@@ -5,6 +5,7 @@
 
 package org.jetbrains.compose.reload.agent
 
+import org.jetbrains.compose.reload.analysis.ClassId
 import org.jetbrains.compose.reload.analysis.ComposeGroupKey
 import org.jetbrains.compose.reload.analysis.Ids
 import org.jetbrains.compose.reload.core.createLogger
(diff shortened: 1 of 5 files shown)


AbdullinAM (changes requested):
Overall, I really like the idea and the implementation. I think this is a good way to solve the issue.

Right now I encountered two issues:
1. ~~Debugging of the target app is broken for me, I can only attach to the agent before devtools started~~ This one is probably related to my IDE being broken
2. (Minor) Because of the change in the startup order, build tool logo detection is broken in devtools. Now it only shows the reload icon

Does this reproduce for you?

AbdullinAM:
https://github.com/user-attachments/assets/5634dee6-ed69-4625-8486-772f5a302718

sellmair:
@AbdullinAM Can you re-test and review the last commit: 
I saw some tests failing with ClassCircularityError. This is due to the change in the startup procedure of the agent. 
If the change does not work, then we should later switch to pre-loading agent classes before adding our transformers.

AbdullinAM:
@sellmair Tested the latest version. Everything works fine, did not see any errors 👍

Issue #195, opened 2025-05-26
Stabilize: Compose API usages


sellmair:
CHR uses some _internal_ APIs from CMP. The usages shall be stabilized and protected from being broken

PR #317, opened 2025-08-07
#195: Stabilise Compose API usages

AbdullinAM:
* Replace access to Compose internal `LocalWindow` with custom detection of the local window
* Completely remove access to `loadStateAndComposeForHotReload` and replace all uses of `retryFailedCompositions` with `resetComposition`
* Replace UI error overlay `retry` button with `restart` button. Motivation:
  * In most cases UI errors are caused by the errors in the source code, so `retryFaieldCompositions` did not help, only thing that can help in this situation is to fix the error and trigger reload
  * In rare cases when UI error is caused by failed composition --- we can't provide any alternatives to `retryFailedComposition` without accessing internal API of Compose. The closes thing we can do --- offer easy way to restart the app


AbdullinAM:
If changes are accepted, I have one more discussion question:
Do we now (1) remove `RetryFailedCompositionRequest` orchestration message completely (causing a breaking change in public API), or (2) keep it and use `resetComposition` stub for it.

I'm personally fine with both, but more leaning to (1), because intuitively I feel like this change will not break anything for most users

pjBooms:
Do we have a real world scenario (or test case) where "Retry failed compositions" helps?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/UIErrorState.kt
```diff
@@ -59,6 +59,6 @@ private fun OrchestrationMessage.UIException.toDescription(): UIErrorDescription
         title = "UI Exception",
         message = message,
         stacktrace = stacktrace,
-        recovery = { OrchestrationMessage.RetryFailedCompositionRequest().sendAsync() }
+        recovery = { OrchestrationMessage.RecompileRequest().sendAsync() }
```
sellmair: This change is not necessarily clear to me:
It seems like you want to recover from the UIException by issueing a full reload?
However, this is effectively a noop in continuous builds and if nothing has changed, IIRC nothing will be done as well?

hot-reload-runtime-api/src/jvmMain/kotlin/org/jetbrains/compose/reload/api.kt
```diff
@@ -32,7 +32,7 @@ public actual fun DevelopmentEntryPoint(child: @Composable () -> Unit) {
     }
 
     /* Forward the call to the runtime-jvm module */
-    org.jetbrains.compose.reload.jvm.DevelopmentEntryPoint(child)
+    org.jetbrains.compose.reload.jvm.DevelopmentEntryPoint(child = child)
```
sellmair, line 35: Isn't this breaking the public version of `DevelopmentEntryPoint` by missing the window?

sellmair (changes requested):

sellmair:
> Do we now (1) remove RetryFailedCompositionRequest orchestration message completely (causing a breaking change in public API), or (2) keep it and use resetComposition stub for it.

IMHO its okay to deprecate it and not care about it. Stubing sounds also fine!

hot-reload-runtime-api/src/jvmMain/kotlin/org/jetbrains/compose/reload/api.kt
sellmair: But do we really care about the public `DevelopmentEntryPoint` actually providing a window?
I think its actually okay as you proposed it here. If necessary we can bring back our own public -api WindowLocal?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/UIErrorState.kt
AbdullinAM: Agree, this change does not make a lot of sense, will revert it

hot-reload-runtime-api/src/jvmMain/kotlin/org/jetbrains/compose/reload/api.kt
AbdullinAM: > This function might only be applicable for non-window-based applications

I was going based on this line form docs. If there is a window-based app --- the window will be detected automatically. If the user app is not window-based --- there is no window to attach to?

PR #318, opened 2025-08-08
Option for a recompiler warmup

AbdullinAM:
* `compose.reload.build.warmup` option to enable warmup of the recompiler at the start of the app
* `ReloadState` is now updated on recompile request --- better/faster visual feedback for the user


sellmair:
cc @singleton11

PR #324, opened 2025-08-11
New DevTools design

AbdullinAM:


sellmair:
Damn, this was quick!
Really like the new error overlay look!

I wonder why the animations on the outlined borders got removed? I thought they were quite satisfying.

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/gradle/GradleRecompilerExtension.kt
```diff
@@ -37,7 +37,7 @@ internal class GradleRecompilerExtension : RecompilerExtension {
         }
 
         /* Side Effect */
-        if (HotReloadEnvironment.gradleBuildContinuous) {
+        if (HotReloadEnvironment.gradleWarmupEnabled || HotReloadEnvironment.gradleBuildContinuous) {
```
sellmair: I wonder why this change shows up here? I thought we merged the warmup?

AbdullinAM: This seems like a strange artefact of the rebase, I will check if everything is OK

AbdullinAM:
> I wonder why the animations on the outlined borders got removed? I thought they were quite satisfying.

The designers suggested that making the background color animated is enough to convey the info, making the borders coloured as well may be "too much" (everything is too red when there is an error). But we can return if we feel that it is better that way 🙁

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/gradle/GradleRecompilerExtension.kt
sellmair: To me it seems like its solved when just rebasing onto latest master!

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtButton.kt
```diff
@@ -27,24 +27,84 @@ import androidx.compose.runtime.remember
...
+
+    // Colors based on state and type
+    val backgroundColor = when {
+        else -> if (isHovered) DtColors.defaultActive else DtColors.applicationBackground
+    }
+
+    val elevation = when {
+        isHovered -> 1.dp
+        else -> 0.5.dp
+    }
```
sellmair: Is there any chance to just use the code from `DtButton`?

sellmair:
Okay how about this: 
We animate colors when something changes, but transition back to 'idle'.
So a 'stable'/'idle' state looks as proposed in the designes, but during reload it also animates and on error we show the color for a second or two?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtButton.kt
AbdullinAM: Refactored `DtButton` so that it can be configurable, and now we can reuse it here 👍

sellmair:
💪 Lets clean up the commit messages and from there, lets go!

PR #329, opened 2025-08-12
Use `--no-parallel` for windows tests in CI

AbdullinAM:
Context:

* `CoroutinesTest::test - switching threads` is flaky on Windows
* The main reason seems to be because it sometimes runs in parallel with other tests, some of them are very resource-intensive (e.g. stress tests)
* By using `--no-parallel` in combination with `-Pchr.tests.sequential` (which was already there, so we kinda already wanted tests to execute one-by-one) we are guaranteed that the test will be executed one-by-one


sellmair:
At this point, I would rather favor rewriting the test or just dropping it! 
Using `-no-parallel` on CI makes me unhappy because it might increase the difference of behaviour on CI vs locally by one more significant option. Also, I suspect even less CPU utilization (which is already super low)

AbdullinAM:
Testing the idea of just increasing the timeout again.

Motivation: I don't want to remove the test completely, because it checks an important feature (nested `withThread` calls) of the coroutines implementation. At the same time, any other attempt to test this feature  may potentially fail with the same problem if the timeout is still small.

PR #333, opened 2025-08-18
Distributed & Atomic Orchestration States

sellmair:
The proposed 'OrchestrationState' facility is supposed to act as a convenient, fast, and safe way of sharing a common 'State' across multiple processes (DevTools, Build, Application, IDE). 

States defined as 'OrchestrationState' can be requested by any process and will be synchronized automatically. 
Updates to any state are atomic using the `update` function. 

```kotlin
fun compile() {
    orchestration.update(ReloadState) { current ->
        current as? ReloadState.Reloading ?: ReloadState.Reloading()
    }
}
```

The orchestration protocol version is bumped to 1.1, which is backwards and forward compatible.
Note that this MR also introduces new test fixtures called 'IsolateTest', which allows writing tests where specific parts of the code will be executed with prior versions of the module's classes.

States are defined in the 'devtools-api' module.
The `ReloadState`, `ReloadCountState` and `WindowsState` were migrated to dogfood the APIs.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,10 +3,10 @@
 ############
 # Version used for publications
 # Will be bumped only right before release
-version=1.0.0-rc01+160
+version=1.0.0-rc01+162
 # Version used to bootstrap the repo
 # Mostly the same as the 'version'.
-bootstrap.version=1.0.0-rc01+160
+bootstrap.version=1.0.0-rc01+162
 compose.reload.autoRuntimeDependenciesEnabled=false
 kotlin.stdlib.default.dependency=false
 ############
--- a/hot-reload-test/sample/build.gradle.kts
+++ b/hot-reload-test/sample/build.gradle.kts
@@ -18,7 +18,7 @@ kotlin {
     jvm()
 
     sourceSets.jvmMain.dependencies {
-        implementation("org.jetbrains.compose.hot-reload:runtime-api:1.0.0-rc01+160")
+        implementation("org.jetbrains.compose.hot-reload:runtime-api:1.0.0-rc01+161")
         implementation(compose.runtime)
     }
 
--- a/hot-reload-test/sample/settings.gradle.kts
+++ b/hot-reload-test/sample/settings.gradle.kts
@@ -5,8 +5,8 @@
 
 pluginManagement {
     plugins {
-        id("org.jetbrains.compose.hot-reload") version "1.0.0-rc01+160"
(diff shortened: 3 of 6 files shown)


hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -37,17 +35,32 @@ internal fun startWindowManager(window: Window): WindowId {
...
+        /* Synchronize windows state with orchestration */
+        launch {
+            windowState.collect { state ->
+                orchestration.update(WindowsState) { current ->
+                    val windows = if (state == null) current.windows - windowId
+                    else current.windows + (windowId to state)
+                    current.copy(windows = windows)
+                }
+            }
+        }
```
sellmair: cc @natanfudge : 
If this change lands, you might want to use the same API to update the WindowState.

hot-reload-core/src/benchmark/kotlin/org/jetbrains/compose/reload/core/TopologicalSortBenchmark.kt
```diff
@@ -6,6 +6,7 @@
 package org.jetbrains.compose.reload.core
 
 import kotlinx.benchmark.*
+import kotlinx.benchmark.State
```
pjBooms, line 9: superflous import

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/State.kt
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+package org.jetbrains.compose.reload.core
+
+import java.util.concurrent.atomic.AtomicReference
+
+
+public interface State<T> {
+    public val value: T
+    public suspend fun collect(collector: suspend (T) -> Unit)
```
pjBooms, line 18: I would add a comment here as otherwise I have to see the implementation to understand  the semantic

hot-reload-core/src/benchmark/kotlin/org/jetbrains/compose/reload/core/TopologicalSortBenchmark.kt
sellmair: I think it's required, otherwise it will use the .core.State we have introduced in this MR

PR #335, opened 2025-08-21
Resolve dirty resources based on ResourceContentHash annotation

pjBooms:
-  remove previous logic


hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
```diff
@@ -56,3 +59,17 @@ internal fun ClassInfo(classNode: ClassNode): ClassInfo? {
...
+private fun getResourceContentHash(fieldNode: FieldNode?): Int? {
+    try {
+        return fieldNode?.invisibleAnnotations?.find { isResourceContentHashAnnotation(it) }?.values[1] as Int?
+    } catch (_: Throwable) {
+        return null
+    }
+}
+
+fun isResourceContentHashAnnotation(it: AnnotationNode): Boolean =
+    it.desc == "Lorg/jetbrains/compose/resources/ResourceContentHash;"
```
sellmair: Can we use ClassId and define it in the universal `Ids` object?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/DirtyResolverExtension.kt
```diff
@@ -1,14 +0,0 @@
-/*
- * Copyright 2024-2025 JetBrains s.r.o. and Compose Hot Reload contributors.
- * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
- */
-
-package org.jetbrains.compose.reload.analysis
-
-import org.jetbrains.compose.reload.core.Context
-
-interface DirtyResolverExtension {
```
sellmair, line 10: This is part of the public API; I expect the CI to fail. I triggered a run for you. Please run tests as well!

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ids.kt
```diff
@@ -192,18 +192,6 @@ object Ids {
...
-        val classId = ClassId("org/jetbrains/compose/resources/StringResourcesKt")
-    }
-
-    object StringArrayResourcesKt {
-        val classId = ClassId("org/jetbrains/compose/resources/StringArrayResourcesKt")
-    }
-
-    object PluralStringResourcesKt {
-        val classId = ClassId("org/jetbrains/compose/resources/PluralStringResourcesKt")
-    }
```
sellmair: Here as well: This is part of the public API already

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/MemberInfo.kt
```diff
@@ -16,7 +16,8 @@ data class FieldInfo internal constructor(
     val fieldId: FieldId,
     val isStatic: Boolean,
     val initialValue: Any?,
-) : MemberInfo() {
+    val resourceContentHash: Int?,
+    ) : MemberInfo() {
```
sellmair: We probably want to reformat this

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/MemberInfo.kt
```diff
@@ -16,7 +16,8 @@ data class FieldInfo internal constructor(
     val fieldId: FieldId,
     val isStatic: Boolean,
     val initialValue: Any?,
-) : MemberInfo() {
+    val resourceContentHash: Int?,
```
sellmair: I think we should call this something more generic. 
The idea here would be the following: 

Right now, we do have this additional hash that we calculate and re-use. Its coming from the resources content. If we want to include more information, instead of introducing more hash fields, we could just mix it into this one hash. 

How about 'extraHash' ?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
```diff
@@ -43,11 +43,11 @@ private fun Context.resolveDirtyRuntimeScopeInfos(
     val dirtyComposeScopes = resolveDirtyComposeScopes(current, redefined) +
         resolveRemovedComposeScopes(current, redefined)
 
-    val dirtyMethods = resolveDirtyMethodsFromExtensionPoints(current, redefined) +
-        resolveDirtyMethods(current, redefined) +
+    val dirtyMethods = resolveDirtyMethods(current, redefined) +
         resolveRemovedMethods(current, redefined)
 
     val dirtyFields = resolveDirtyFields(current, redefined) +
+        resolveDirtyComposeResourceAccessors(current, redefined) +
```
sellmair: How about incorporating this check into resolveDirtyFields?
I do think a single

```
previousField.initialValue != redefinedField.initialValue ||   previousField.resourceContentHash != redefinedField.resourceContentHash
```

would simply do the trick

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
```diff
@@ -293,7 +300,7 @@ private fun resolveTransitivelyDirty(
                 ?: current.classIndex[element.memberId.classId]
 
             classInfo?.fields?.forEach { (fieldId, field) ->
-                if (field.isStatic) {
+                if (field.isStatic && field.resourceContentHash == null) {
```
sellmair: Upon first reading, this is not clear to me: 
Can you provide some insights on when this check matters?

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/gradle/ComposeHotRun.kt
```diff
@@ -18,6 +18,11 @@ import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 
 @UntrackedTask(because = "This task should always run")
 sealed class AbstractComposeHotRun : JavaExec(), ComposeHotReloadRunTask {
+
+    init {
+        System.setProperty("compose.resources.generate.ResourceContentHash.annotation", "true")
+    }
```
sellmair: This is a blocker: 
We cannot just set a global System property in the init of a task. 
When reading the previous MR I misunderstood this as a System property that lives in isolation. 

Please configure the corresponding generation tasks properly.

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ResourcesTests.kt
```diff
@@ -279,6 +273,57 @@ class ResourcesTests {
...
 
+    @HotReloadTest
+    @QuickTest
+    fun `change text resource in inner scope`(fixture: HotReloadTestFixture) = fixture.runTest {
+        val testResource = testStringResourceFile()
+        val xmlBefore = stringsXmlResource(
+            """
+                     <string name="outer">outer</string>
+                     <string name="inner">inner before</string>
+                """
```
sellmair: We probably want to check the formatting of the String templates in this file again.

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ResourcesTests.kt
```diff
@@ -279,6 +273,57 @@ class ResourcesTests {
...
+                            TestText(stringResource(Res.string.inner) + " (" + state + ")")
+                        }
+                    }
+                }
+            }
+        """.trimIndent()
+        )
+        checkScreenshot("before")
+        fixture.sendTestEvent()
+        fixture.checkScreenshot("before-1")
```
sellmair, line 318: Nice test idea! Cool to see that it keeps the state!

properties.yaml
```diff
@@ -340,15 +340,3 @@ IsHotReloadActive:
...
-  key: compose.reload.resourcesDirtyResolverEnabled
-  type: boolean
-  default: "false"
-  target: [application, build, devtools]
-  isDelicateApi: true
-  documentation: |
-    Enables/Disables automatically re-composing scopes using resources
-    See: https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-multiplatform-resources.html
-    See: https://github.com/JetBrains/compose-hot-reload/issues/151
-    Feature is still in development and subject to change in the future.
```
sellmair: Likely also part of the api already

sellmair (changes requested):

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/DirtyResolverExtension.kt
pjBooms: May we remove  something from the public API at this point?

properties.yaml
pjBooms: I have not found any .api file for it.

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
pjBooms: The code marks all fields dirty if a class initializer changes. If we talk about resource accessors it means:

-- if we change one text resource, the class initializer of a resource accessor class will change (due to internal implementation)
-- the code will mark all resource accessors as dirty even if only one has changed
-- then we invalidate more scopes that it is actually required

Filtering resource accessor fields from the picture fixes the problem.

In general the code is too conservative. Ideally, it is better to analyze which static fileds are actually changed in the changed initializer to mark them dirty (not all of them).

My change is also safe because we already mark changed resource accessors as dirty (and not more).

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
pjBooms: yes, done

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/DirtyResolverExtension.kt
pjBooms: Finally I marked it as deprecated

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ids.kt
pjBooms: Deprecated now. Though I can't imagine any meaningful usage of this "public API". I believe all other ids should be declared internal. But if you see a meaningful usage, please share.

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/MemberInfo.kt
pjBooms: fixed

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/MemberInfo.kt
pjBooms: Called it as "changeIndicatorHash" to reflect semantic.

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
pjBooms: Done! I thought about it in the beginning but was not sure if it is ok to mix it. However after we renamed resourceContentHash to a neutral name, now it seems to be ok to mix

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/gradle/ComposeHotRun.kt
pjBooms: Fixed In CMP (introduced a task property for this) and here! (I thought that we agreed with `System.property` way because I communicated about it with you, but it seems we misunderstood each other. And yes, It caught my eye too, I just was not sure how to rework this as `System.property` was a negotiated way in my mind and landed to CMP)

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
```diff
@@ -43,7 +45,8 @@ internal fun ClassInfo(classNode: ClassNode): ClassInfo? {
         FieldId(classNode, fieldNode) to FieldInfo(
             fieldId = FieldId(classNode, fieldNode),
             isStatic = fieldNode.access and (Opcodes.ACC_STATIC) != 0,
-            initialValue = fieldNode.value
+            initialValue = fieldNode.value,
+            changeIndicatorHash = getResourceContentHash(fieldNode)
```
sellmair: Initial thought here:
When I read this, I might think 'oh if this hash is changed, then the field is dirty', but it is more like an 'additional' hash: It might still be dirty, even if this 'additional' hash is unchanged (e.g. by the initial value changing): 

Do you think we can help with intuition by calling it 'additionalHash', or 'extraHash' + some documentation?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
```diff
@@ -293,7 +284,7 @@ private fun resolveTransitivelyDirty(
                 ?: current.classIndex[element.memberId.classId]
 
             classInfo?.fields?.forEach { (fieldId, field) ->
-                if (field.isStatic) {
+                if (field.isStatic && field.changeIndicatorHash == null) {
```
sellmair: Note: I still wonder why this condition is added

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/gradle/hotRunTasks.kt
```diff
@@ -63,6 +64,17 @@ internal val Project.hotRunTasks: Future<Collection<TaskProvider<out AbstractCom
         task.args("--className", task.className.string(), "--funName", task.funName.string())
     }
 
+
+    tasks.matching { task ->
```
sellmair: There are three issues here:

1st: I think configuring external compose multiplatform tasks sh ould not be done here, but next to the plugin (we can do this in pair programming)

2nd: The tasks.matching {} will effectively be a huge performance issue, as it forces all tasks to materialize (breaks laziness)

3rd: I am not sure if 'matching by name' is actually a good idea: Can't we just use the type?

sellmair (changes requested):

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/gradle/ComposeHotRun.kt
sellmair: I was under the impression that the System property from the CMP is passed to an isolated process/task. 
A task property is an OK way to communicate; The system property in the Gradle process was not

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ResourcesTests.kt
pjBooms: fixed

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
pjBooms: https://github.com/JetBrains/compose-hot-reload/pull/335#discussion_r2298795969

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
pjBooms: How about additionalChangeIndicatorHash + doc?

When I see `additionalHash`, I have a question "hash of what?"

sellmair: > How about additionalChangeIndicatorHash + doc?

I like it 😎

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
sellmair: Okay, in this case, would it be possible to move the filtering to 'Phase 1' instead? Similar to how we handle `ComposableSingleton`?

pjBooms: I do not think it is good idea to filter the whole class initializer: we may have both fields that are resource accessors and ordinary fields in the class. On the other hand we can't filter just a field, cause it can be added to the queue from two sources: when it is a resource accessor and changed, and when it is a part of a dirty class initiliazer. And we can't differentiate this cases in the phase 1.

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
pjBooms: Done

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/gradle/hotRunTasks.kt
pjBooms: Introduced [public API](https://github.com/JetBrains/compose-multiplatform/pull/5402) in CMP Gradle Plugin for this and used from here.

hot-reload-analysis/api/hot-reload-analysis.api
```diff
@@ -157,7 +157,9 @@ public final class org/jetbrains/compose/reload/analysis/FieldInfo : org/jetbrai
 	public final fun component1 ()Lorg/jetbrains/compose/reload/analysis/FieldId;
 	public final fun component2 ()Z
 	public final fun component3 ()Ljava/lang/Object;
+	public final fun component4 ()Ljava/lang/Integer;
 	public fun equals (Ljava/lang/Object;)Z
+	public final fun getChangeIndicatorHash ()Ljava/lang/Integer;
```
sellmair: I wonder why this is called 'getChangeIndicatorHash' while the property is called 'additionalChangeIndicatorHash'?

sellmair: I think we need to update the apiDump here 🙇

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/resolveDirty.kt
sellmair: Hmm, I can totally see that we will have to think about that later. For now let's go with the rule:
If it works, it works. If we see an issue, then we will adjust the code to work for both cases

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/gradle/hotRunTasks.kt
```diff
@@ -63,6 +63,20 @@ internal val Project.hotRunTasks: Future<Collection<TaskProvider<out AbstractCom
         task.args("--className", task.className.string(), "--funName", task.funName.string())
     }
 
+
+    /* Configure resource accessors generation of Compose Gradle Plugin */
+    try {
```
sellmair: Code LGTM, but lets move that into another source file!
The 'hotRunTasks.kt' is supposed to contain code configuring the run tasks. 
This code configures the resources task. 

How about creating a file like 'composeMultiplatform.kt'  which contains this and future code where we call into cmp?

pjBooms: done

hot-reload-analysis/api/hot-reload-analysis.api
pjBooms: API is updated

PR #336, opened 2025-08-21
Hide devtools ui on shutdown request

AbdullinAM:
Added `DtLifecycleState` that holds `isActive` property. It flips on `ShutdownRequest`, and the devtools ui is hidden accordingly.

Notes:
* `appliсationScope.cancel()` did not speed up the ui shutdown, so I opted to just return
* decided to use simple boolean property, we can easily extend it in the future if we want. Considering this is devtools internal api --- will not cause any compatibility problems

### DIFF
--- a/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/DtLifecycleState.kt
+++ b/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/DtLifecycleState.kt
@@ -1,0 +1,16 @@
+/*
+ * Copyright 2024-2025 JetBrains s.r.o. and Compose Hot Reload contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.compose.devtools.states
+
+import io.sellmair.evas.State
+
+data class DtLifecycleState(
+    val isActive: Boolean,
+) : State {
+    companion object Key : State.Key<DtLifecycleState> {
+        override val default: DtLifecycleState = DtLifecycleState(isActive = true)
+    }
+}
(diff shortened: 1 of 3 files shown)


hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/DtLifecycleState.kt
```diff
@@ -0,0 +1,34 @@
...
+    orchestration: OrchestrationHandle = org.jetbrains.compose.devtools.orchestration
+) = launchState(DtLifecycleState.Key) {
+    orchestration.asFlow().collect { message ->
+        if (message is OrchestrationMessage.ShutdownRequest) {
+            DtLifecycleState.update { state ->
+                state.copy(isActive = false)
+            }
+        }
+    }
+}
```
sellmair: Do you think it would be simpler if we would just write 
in the shutdown() function?

```
applicationScope.coroutineContext.statesOrThrow.setState()
```

I am just unsure if we can always realy on the ShtudownRequest message being flushed?

AbdullinAM: I think in general `ShtudownRequest` should be pretty reliable, because otherwise the main app wouldn't close? But overall I agree, setting it manually would be more safe

sellmair: I mean, in case of the main app exiting, we might not get a shutdown request. We receive an 'ApplicationWindowGone' event, therefore I think it would be more universal if the shutdown function just switches the state?

AbdullinAM: Yes, I agree. I updated the implementation

sellmair (approved):

PR #344, opened 2025-08-27
Orchestration: Lenient packageType & opaque orchestration messages

sellmair:
Lenient packageType:
Previously, the OrchestrationIO would fail when a package with an unknown package type was received. This might happen if a new package type is introduced and an old client/server sees this package type. In order to simplify keeping forward compatibility, the IO will just skip packages of unknown type.

Opaque Orchestration Messages:
Servers which might not be able to deserialize a provided message (e.g. because of this message being introduced in a newer version of the protocol) are supposed to still dispatch this message to clients. This shall also help in providing forward compatibility. Any such message is considered 'opaque' to the server (aka the server does not know what message this is). Such opaque messages will still be emitted and distributed to all connected clients.


pjBooms (commented):
How do we test it?

sellmair:
Backwards and forward compatibility tests will be used to test this. 
Right now, those changes are here to help keep them green in the future.
The behaviour is unchanged, right now. If we would like to add messages to the orchestration, the compatibility tests might break and we will be able to use those tools and changes to keep them green.

Automatically testing this scenario proactively is hard because we do not know upfront what kind of change we would like to do!

pjBooms:
>Automatically testing this scenario proactively is hard because we do not know upfront what kind of change we would like to do!

>Servers which might not be able to deserialize a provided message (e.g. because of this message being introduced in a >newer version of the protocol) are supposed to still dispatch this message to clients

Can't we connect a testing client that does not change a protocol but sends at some point an unknown message for testing purposes?

sellmair:
Good idea! I thought about something similar, but then I expected issues because we _cannot_ simply add a message type for tests. Let me try to find a way here (e.g. by inline compiling a class) 

If it gets too hard, then I would just leave it, as the tests are 'does not break any behaviour' and should help in the future

PR #347, opened 2025-08-28
DevTools: Provide 'WindowExceptionHandler' logging to the orchestration instead of crashing devtools

sellmair:
…on instead of crashing devtools


sellmair:
image
Example log

PR #350, opened 2025-09-02
Improve GCC compliance for orchestration port & pidFile based configurations

### LINKED ISSUE #340: Recompile Warmup: Calculating task graph as configuration cache cannot be reused because system property 'compose.reload.orchestration.port' has changed

sellmair:
Fixes #340


sellmair:
Let me check! I only ran a subset of tests locally and saw a green CI run. But now it seems to have a failure on CI as well

sellmair:
Fixed: Seems like 'allJvmArgs' does not return 'all jvm args' and we need to also provide 'jvmArgs' (as before)
https://github.com/JetBrains/compose-hot-reload/compare/056321c6f2df249747a521f35da03ffe7eca930d..91d2031210802e902c3c7983dcd0dac77221700c

PR #354, opened 2025-09-04
Update WindowsState immediately on the window state/position change

AbdullinAM:
Motivation: previous solution did not properly update `WindowsState` on exit, causing the devtools window to stay visible for too long.

On exit `broadcastGone` successfully sends `ApplicationWindowGone` and updates the `MutableStateFlow`. Before introducing orchestration states, `WindowsState` was managed on the devtools side and it was updated upon `ApplicationWindowPositioned/ApplicationWindowGone` messages. 

Now, these messages are (basically) not used and the `WindowsState` is managed by a coroutine in `startWindowManager`, that collects the changes on the `MutableStateFlow`. However, when the user closes the application, the coroutine does not have time to update the `WindowsState` as the process is shutting down. Therefore, devtools do not receive the update of the `WindowsState`.

Solution: get rid of the coroutine and `MutableStateFlow` and explicitly update the `WindowsState` in `broadcastActiveState` and `broadcastGone`.


sellmair (approved):
LGTM, however, I am still slightly concerned that we won't be able to update the window in certain situations (race?); Let's wait for issues and react to them when they appear.

PR #355, opened 2025-09-04
New reload status overlay

AbdullinAM:


samples/counter/settings.gradle.kts
```diff
@@ -18,7 +18,7 @@ pluginManagement {
     plugins {
         kotlin("multiplatform") version "2.2.0"
         kotlin("plugin.compose") version "2.2.0"
-        id("org.jetbrains.compose") version "1.8.2"
+        id("org.jetbrains.compose") version "1.9.0-rc01"
```
sellmair: QQ, did you use any API only available in a newer version of compose?

AbdullinAM: No, but this was done for testing. Currently `ReloadOverlay` is compiled with compose `1.9.0-rc01`, and if the target app has lower version of compose, we fail with the `NoSuchMethodException`:
```
Error was captured in composition while live edit was enabled.
java.lang.NoSuchMethodError: 'long androidx.compose.runtime.ComposablesKt.getCurrentCompositeKeyHashCode(androidx.compose.runtime.Composer, int)'
	at org.jetbrains.compose.reload.jvm.ReloadOverlayKt.ReloadOverlay(ReloadOverlay.kt:270)
	at org.jetbrains.compose.reload.jvm.JvmDevelopmentEntryPoint.DevelopmentEntryPoint$lambda$11(DevelopmentEntryPoint.kt:78)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
	at org.jetbrains.compose.reload.jvm.JvmDevelopmentEntryPoint.DevelopmentEntryPoint$lambda$12(DevelopmentEntryPoint.kt:102)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
	at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider(CompositionLocal.kt:384)
```

I suspect that there was ABI change between `1.8.2` and `1.9.0-rc01` that causes the issue. This is the reason the tests are failing now on CI

I'm currently trying to come up with a workaround for this

AbdullinAM:
Still trying to solve the issue with the exception

PR #357, opened 2025-09-07
Add information about source location to label tokens.

avan1235:
The source location starts being exposed via `InvalidatedComposeGroupMessage` after invalidation of hot reload


avan1235:
I'm aware these changes break an API of `org/jetbrains/compose/reload/analysis/InstructionToken$LabelToken` in not backwards compatible manner, but maybe that's a sign not to use data classes for such exposed types before stable, and allow for this breakage this time, wdyt?

I don't see a way of keeping backward compatibility for this class, since it's about synthetic `copy` method and some maker for the default constructor. I saw other libraries not using data classes because of these reasons.

When it comes to implementation detail, I decided to extend the `LabelToken` since there's a relation afaiu that each line number has some label, while some label can have no line numbers. This seemed to be least intrusive change that allows collecting interesting for me instructions with line numbers. I also tried to tokenize them separately, but this introduces unnecessary noise of mapping/relating them with proper labels, and line is just not needed as separate abstraction there from my perspective.

sellmair:
I read the changes initially, and it looks really good to me 👍 

Regarding binary compatibility: 
I think the `LabelToken` can use ja `@JvmOverloads` on the constructor and a manual implementation of the previous copy function:

```kotlin
    data class LabelToken @JvmOverloads constructor(
        val labelInsn: LabelNode,
        val lineNumberInsn: LineNumberNode? = null,
    ) : InstructionToken() {

        fun copy(labelInsn: LabelNode = this.labelInsn): LabelToken {
            return LabelToken(labelInsn, lineNumberInsn)
        }

        override val instructions: List<AbstractInsnNode> = listOfNotNull(labelInsn, lineNumberInsn)
        override fun toString(): String {
            return "LabelToken(label=${labelInsn.label}, lineNumber=${lineNumberInsn?.line})"
        }
    }
```

From there I see only additions in the api dump. 
image

sellmair:
I changed your target branch to 'staging'. We usually merge into this branch first 👍

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
```diff
@@ -297,6 +297,39 @@ internal constructor() : OrchestrationPackage(), Serializable {
         }
     }
 
+    public data class InvalidatedComposeGroupMessage(
```
sellmair, line 300: Adding a new Serializable orchestration message will break old servers/clients. 
Let's add a Backwards & Forward compatibility test (see `ServerBackwardCompatibilityTest` & `ServerForwardCompatibilityTest`)

I expect older beta versions not to handle seeing this message gracefully. 


However, this is not a huge problem: 
I think adding a message can bump the protocol version. 
Maybe we can then add something like 'since protocol version' and just not send messages to clients with too low a protocol version 🤔 

However, if you find some other approach, feel free to propose it. 
Note: Since beta06, servers and clients will just handle such cases gracefully. The server will even distribute the message, despite not knowing its content.

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
```diff
@@ -297,6 +297,39 @@ internal constructor() : OrchestrationPackage(), Serializable {
...
+            val sourceFile: String?,
+            val firstLineNumber: Int?,
+        ) : Serializable {
+            internal companion object {
+                @Suppress("unused")
+                internal const val serialVersionUID: Long = 0L
+            }
+
+            public enum class ScopeType : Serializable {
+                Method, RestartGroup, ReplaceGroup, SourceInformationMarker;
```
sellmair, line 323: Just wondering: 
Is the ScopeType used in the IDE?

sellmair (changes requested):
Looks really good!
Two open TODOs: 

1st) Check if you think applying the changes to LabelNode for compatibility works for you

2) Test (and restore) the backwards compatibility of the new orchestration message

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
avan1235: not yet, I was about asking you what those exactly are and if we could use them. I think currently used one was only `Method` one. But I remembered that you mentioned this being part of expected orchestration message I think, and it just felt right to include it (other parts like method descriptor is not used yet too, but will be for the fallback mode)

sellmair: Agree ✅

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
avan1235: Thanks for fixing the part with checking compatibility based on version, current approach lgtm

AbdullinAM (approved):

PR #362, opened 2025-09-10
Update readme

sarahhaggarty:
This PR updates the readme file.


README.md
```diff
@@ -7,102 +7,167 @@
...
-No restarts required.
-Compose Hot Reload runs your application on the JetBrains Runtime 
-and intelligently reloads your code whenever it is changed.
+With Compose Hot Reload, you can make UI code changes in a Compose Multiplatform app and see the results instantly, without needing to restart.
+The JetBrains Runtime intelligently reloads your code whenever it changes.
 
-## Getting Started
+> [!IMPORTANT]  
+> Compose Hot Reload only works if you have a desktop target in your multiplatform project. We're exploring adding support for 
+> other targets in the future.
```
sellmair: What do you think about switching from 'only works' to something like "requires a JVM target".

README.md
```diff
@@ -7,102 +7,167 @@
...
+
+## Prerequisites
+
+- Kotlin 2.1.20 or higher.
+- Compose compiler 2.1.20 or higher.
+- JetBrains Runtime.
+
+## Set up your project
```
sellmair, line 30: Wow, thats a quite detailed description!
However, I am not sure if we want to maintain a generic 'how to create a project' guide here, especially since the project setup guide seems rather KMP focused'. Do we this information hosted somewhere else as well? We could maybe just point there?

README.md
```diff
@@ -7,102 +7,167 @@
...
+
+The Compose Hot Reload plugin automatically creates the following tasks to launch the application in 'hot reload mode':
+
+- `:hotRunJvm`: For multiplatform projects. The async alternative is `:hotRunJvmAsync`.
+- `:hotRun`: For Kotlin/JVM projects. The async alternative `:hotRunAsync`.
+
+You can run these Gradle tasks from the command line:
+
+```shell
+./gradlew :app:hotRunDesktop
```
sellmair: Probably slipped through. Should likely be `:app:hotRunJvm`

README.md
```diff
@@ -7,102 +7,167 @@
...
+
+5. Click the **Sync Gradle Changes** button to synchronize Gradle files:
+
+    <img alt="Synchronize Gradle files" src="./readme-assets/gradle-sync.png" width="50">
+
+## Use Compose Hot Reload
+
+You can run your application with Compose Hot Reload from inside your IDE or from the CLI by using Gradle tasks.
+
+### From the IDE
```
sellmair, line 83: The given screenshot requires the KMP IDE plugin, which is not yet available on most platforms (linux, windows). So I am not sure if we should explain this?

sellmair (changes requested):
~

README.md
sarahhaggarty: This is taken from the [KMP quickstart guide](https://www.jetbrains.com/help/kotlin-multiplatform-dev/quickstart.html#set-up-the-environment). So, if you prefer, we can link to this content. My intention was to try and keep the user in one place so they didn't have to navigate away to complete their setup. But I understand your concern about maintaining this content :)

README.md
sarahhaggarty: Sure! I can add a note to this section explaining that it is only possible for MacOS users currently.

README.md
sarahhaggarty: I updated this and the FAQ question which talks about a desktop target (I'm developing an Android-only app without Kotlin Mutliplatform. Can I use Compose Hot Reload?). Can you check that part is OK?

README.md
sellmair: Got it! Noble goal! However, I feel like we might never satisfy this goal from a project like 'hot reload'. We might have now explained how to setup a KMP project but there are users which do want to use it for IntelliJ plugins or generic desktop apps. Linking to the guide seems like a really good tradeoff: You could find the information and it will be always up-to-date

README.md
sellmair: At this point I trust your gut feeling: 
On one hand saying that a core feature like 'running from gutter' only works on MacOS sounds a little discouraging.  

It *does* work on other platforms as well, but it looks different. Explaining this sufficiently here seems hard.
Please think about it carefully, I trust your decision 🙇

README.md
sarahhaggarty: OK! I replaced the detailled instructions with a link to the KMP quickstart guide :)

PR #369, opened 2025-09-16
Implement public API for custom effects

sellmair:


hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/effects/BorderGlowEffect.kt
```diff
@@ -3,31 +3,121 @@
...
+
+            /* Transition the displayed color */
+            launch {
+                if (!isVisibleBefore) currentColor.snapTo(targetColor)
+                else currentColor.animateTo(targetColor, animationSpec = tween(500))
+            }
+
+            /* Display 'pumping' effect when state was OK */
+            launch {
+                if (state is ReloadState.Ok) {
```
AbdullinAM, line 88: New effect animations are now de-synced from the devtools window animations. Should we try to keep them synced for the built-in effects?

AbdullinAM: New effects also seem to have faster animations. I guess it comes to the personal preference: I would prefer them to be a little slower. Several times I noticed that the border glow have already dissapeared before I noticed it. WDYT?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/effects/BorderGlowEffect.kt
```diff
@@ -3,31 +3,121 @@
...
+                    currentScale.animateTo(
+                        3f, tween(durationMillis = 250, easing = FastOutSlowInEasing)
```
AbdullinAM: Same thing with the preference: if the whole border glow effect is a bit thinner now, maybe we scale only up to `2f` here?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/HotReloadEffect.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.jetbrains.compose.reload.jvm.HotReloadEffect.OverlayEffect
+import java.util.ServiceLoader
+
+/**
+ * Provides effects rendered in the user application.
+ * A given implementation class has to be registered as Service in
+ * `META-INF/services/org.jetbrains.compose.reload.jvm.HotReloadEffect`
+ */
+@ExperimentalHotReloadApi
+public sealed interface HotReloadEffect {
```
AbdullinAM: Did you consider the idea of moving these API declarations to `hot-reload-runtime-api` module? I don't have a particular problem with them being here, but as an end user I would expect all the APIs to be declared there. 

Now the users will have to manually add `hot-reload-runtime-jvm` as a compile-time dependency to their project.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/HotReloadEffect.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * See [HotReloadEffect.priority]
+     */
+    @ExperimentalHotReloadApi
+    @JvmInline
+    public value class Priority(private val value: Int) : Comparable<Priority> {
```
AbdullinAM: Can we add tests that check the sideloading, priorities and ordinals?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/HotReloadEffect.kt
```diff
@@ -0,0 +1,142 @@
...
+        @ExperimentalHotReloadApi
+        public companion object {
+            public val first: Ordinal = Ordinal(0)
+            public val medium: Ordinal = Ordinal(128)
+            public val last: Ordinal = Ordinal(Integer.MAX_VALUE / 2)
+        }
+
+    }
+
+    /**
```
AbdullinAM: I love the idea of providing both priority and ordinal for managing the reload effects🔥! This gives enough flexibility to create whatever you want

Slight concern: opening an API for creating custom values of `Priority` and `Ordinal` might cause some problems in the future if we will want to change something about built-in values/default effects in the future

AbdullinAM: I don't expect users to build 3--5+ effects at the same time, so maybe providing a set number of 3 (or 5) default values for priorities and ordinals will be enough?

Especially considering that if someone really wants it, they can overcome these limitations by just combining the effects in a single class

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/effects/BorderGlowEffect.kt
sellmair: Good question; I think we should meet and just fine tune params together, this will be the easiest. 
Once we found params we both like, then we could share some in the devtools-api and expose them for the IDE as well?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/HotReloadEffect.kt
sellmair: Good point; I think we could do that!

AbdullinAM (approved):
Nice! Really love the new look and the code is super clean as well!🤩

AbdullinAM (approved):

Issue #370, opened 2025-09-16
Reload overlay breaks autosized windows


Nohus:
The snippet below displays a 400x400 window. The window width is set to 400, and the height will automatically fit the content, which is 400 high.

However, when run with CHR with the overlay enabled, the window height expands to fill the entire screen.

```Kotlin
fun main() {
    val size = DpSize(400.dp, Dp.Unspecified)
    singleWindowApplication(state = WindowState(size = size)) {
        Box(
            contentAlignment = Alignment.Center,
            modifier = Modifier.height(400.dp)
        ) {
            Text("Size: $size")
        }
    }
}
```

**Expected behavior**

CHR does not break window sizing.

**Environment**
 - OS: Kubuntu 24.04
 - Kotlin version: `2.2.0`
 - Compose Multiplatform version: `1.8.0`
 - Compose Hot Reload version: `1.0.0-beta07`

AbdullinAM:
Fixed in #369, added tests in #372

PR #372, opened 2025-09-16
Add tests with windows of unspecified size for reload effects

AbdullinAM:
#370 itself was fixed in #369, this PR adds tests that check the fix


tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ReloadEffectsTest.kt
```diff
@@ -44,6 +41,69 @@ class ReloadEffectsTest {
...
+    fun `test - reload effects on window with unspecified size`(fixture: HotReloadTestFixture) =
+        testReloadEffects(
+            fixture, """
+            import androidx.compose.foundation.layout.*
+            import androidx.compose.ui.unit.sp
+            import androidx.compose.ui.window.*
+            import org.jetbrains.compose.reload.test.*
+            
+            fun main() {
+                screenshotTestApplication(width = 0, height = 0) {
```
sellmair: My gut is telling me, that using -1 and checking for >= 0 is more intuitive / conventional. What does your gut tell you?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/runHeadless.kt
```diff
@@ -261,6 +265,36 @@ public fun runHeadlessApplicationBlocking(
     }
 }
 
+/**
+ * Assumes there are no animations that increase in size
+ */
+private fun computeSceneSize(
```
sellmair, line 271: Very cool 😍

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/ReloadEffectsTest.kt
AbdullinAM: Yes, I would agree!

AbdullinAM: However, window size 0 will currently break the setup: we can't create a compose scene of width or height 0, it will throw an exception. So I suggest the best of both worlds: using `-1` in tests, but still checking for `> 0`. WDYT?

PR #373, opened 2025-09-16
Fixed glow shader

### LINKED ISSUE #368: Reload overlay makes it hard to see changes
In a recent version a new overlay was added, where the whole application window turns orange and green as a reload is in progress, covering the application content. This makes it hard to see the change taking place.

For example, previously you could change a padding from 8.dp to 10.dp and see the content move before your eyes to immediately see how it affected the layout. Now the layout gets hidden for the few seconds and then the new layout is shown, so you can no longer visually see what changed / moved around.

It would be better if the overlay was only a window border effect for example, not an overlay for the entire window that hides the content. Or ideally provide a `ReloadIndicator()` composable that can be put wherever the developer wants, like a status bar if the application has one.

I am aware the overlay can be turned off with `compose.reload.reloadOverlayEnabled`; I don't want to do that because the feedback it's giving is very useful – it's just that I would like to get that feedback without my application being completely covered up.

Nohus:
Fixes #368

The clamp method was used incorrectly in the GLSL shader.

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/hot-reload-runtime-jvm/src/main/resources/shaders/glow.glsl
+++ b/hot-reload-runtime-jvm/src/main/resources/shaders/glow.glsl
@@ -150,8 +150,8 @@ half4 renderGlow(vec2 position, vec2 halfSize) {
 
     // Combine masks and color
     float strength = clamp(rimMask * 0.9 + haloMask * 0.7, 0.0, 1.0);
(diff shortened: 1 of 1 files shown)


sellmair:
Thank you very much 🙏

PR #377, opened 2025-09-18
Fix potential deadlock in `WorkerThread`

AbdullinAM:
Is is possible to receive a situation when the mai loop of `WorkerThread::run` is stuck on `queue.take()` on shutdown if the decrement of `pendingDispatches` happened after the work has been taken from the queue. 

Now we add an empty work to the `WorkingThread` queue if the `pendingDispatches` has reached `Int.MIN_VALUE` to awaken the `run` loop

### DIFF
--- a/hot-reload-core/src/benchmark/kotlin/org/jetbrains/compose/reload/core/WorkerThreadBenchmark.kt
+++ b/hot-reload-core/src/benchmark/kotlin/org/jetbrains/compose/reload/core/WorkerThreadBenchmark.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2024-2025 JetBrains s.r.o. and Compose Hot Reload contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.compose.reload.core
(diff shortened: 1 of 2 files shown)


AbdullinAM:
@sellmair 

I tested the throughput of `WorkerThread` using three different configurations:
* check for `if (element.action == null)`
* check for `if (element is Work.Shutdown)` when `Shutdown` is an object
* check for `if (element is Work.Shutdown)` when `Shutdown` is a class instance

Results:
```
benchmark summary:
Benchmark                                    Mode  Cnt       Score      Error  Units
WorkerThreadBenchmark.invokeWorkerNull      thrpt    5  111589,446 ±  1669,237  ops/s // null check
WorkerThreadBenchmark.invokeWorkerObject    thrpt    5  111038,508 ±  2336,341  ops/s // Shutdown is an object
WorkerThreadBenchmark.invokeWorkerClass     thrpt    5  112582,985 ±  3737,034  ops/s // Shutodwn is a class instance
```

The difference is basically negligible, so I decided to stay with the `object` implementation, because it looks nicer :)

sellmair:
LGTM!

PR #382, opened 2025-09-30
update: clarify the Java version requirement related to JBR

zamulla:
Changes mirroring the doc update: https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/522/files


README.md
```diff
@@ -25,8 +25,9 @@ The JetBrains Runtime intelligently reloads your code whenever it changes.
 
 - Kotlin 2.1.20 or higher.
 - Compose compiler 2.1.20 or higher.
-- Compose 1.8.2 or higher
-- JetBrains Runtime.
+- Compose Multiplatform 1.8.2
```
zamulla, line 28: Is that right? "Compose" by itself seems ambiguous.

AbdullinAM: Seems lgtm! Thanks!

README.md
```diff
@@ -25,8 +25,9 @@ The JetBrains Runtime intelligently reloads your code whenever it changes.
...
+- [JetBrains Runtime](https://github.com/JetBrains/JetBrainsRuntime).
+  To be compatible with JetBrains Runtime, your project needs to use Java 21 or older.
```
AbdullinAM: I would maybe rephrase that to "To be compatible with JetBrains Runtime, your project needs to be compiled with jdk 21 or older"? WDYT?

zamulla: Yeah, this is probably more precise, thanks. Updated 👍

README.md
```diff
@@ -25,8 +25,9 @@ The JetBrains Runtime intelligently reloads your code whenever it changes.
 
 - Kotlin 2.1.20 or higher.
 - Compose compiler 2.1.20 or higher.
-- Compose 1.8.2 or higher
-- JetBrains Runtime.
+- Compose Multiplatform 1.8.2
+- [JetBrains Runtime](https://github.com/JetBrains/JetBrainsRuntime).
+  To be compatible with JetBrains Runtime, your project needs to be compiled with JDK 21 or older.
```
sellmair: It requires targetCompatibility with 21 or older, but a higher JDK can be used to produce such compatible files.

zamulla: Makes sense, thanks. I linked the Gradle docs so that users could choose the method for targeting — they say that `targetCompatibility` specifically is "not generally advised", but I guess that your mileage may vary.

sellmair: Oh, thats interesting, I did not know that they do not advise it. Do you happen to know what might go wrong with using it?

zamulla: I don't have any firsthand knowledge, but this seems relevant: https://docs.gradle.org/current/userguide/toolchains.html#sec:source-target-toolchain

> This does not guarantee the correct JDK is used and may cause issues when APIs have been backported to older Java versions.
> You should only use this method in cases where you need backward compatibility but cannot use toolchains.

sellmair: Oh got it; On the other hand, using toolchains instead is also said to be a 'bad practice'. So in any case its all danger and pits, probably OK stay un-opinionated.

sellmair:
I removed myself as reviewer: @AbdullinAM is responsible 👍

README.md
zamulla: Yeah, I agree. So I linked the docs rather than highlighting any particular solution, whatever you use for compatibility — do that :)

AbdullinAM:
I think that current wording is good 👍

AbdullinAM:
@zamulla Let's squash the commits and then we can merge

hmy65:
https://github.com/JetBrains/JetBrainsRuntime/releases/tag/jbr-release-25b176.2
Please make it compatible with JBR 25 thanks. @AbdullinAM
```
LinkageError
	java.lang.UnsupportedClassVersionError: xxx/xxx/xxx/XxxKt has been compiled by a more recent version of the Java Runtime (class file version 68.0), this version of the Java Runtime only recognizes class file versions up to 65.0
```

PR #383, opened 2025-09-30
Clarify run options for Compose Hot Reload from the IDE

sarahhaggarty:
This PR updates the readme file to give more refined instructions how to run Compose Hot Reload from the IDE.


README.md
```diff
@@ -83,12 +83,10 @@ You can run your application with Compose Hot Reload from inside your IDE or fro
...
 ### From the IDE
 
-> [!IMPORTANT]  
-> Currently, only MacOS is supported for running Compose Hot Reload from the gutter in your IDE. We're working on adding support for Windows and Linux.
-
-In IntelliJ IDEA or Android Studio, in the gutter, click the **Run** icon of your main function. Select **Run 'composeApp [hotRunJvm]' with Compose Hot Reload (Beta)**.
+In IntelliJ IDEA or Android Studio, in the gutter, click the **Run** icon <img alt="Run main function" src="./readme-assets/run.png" width="12"> of your main function:
 
-<img alt="Run Compose Hot Reload from gutter" src="./readme-assets/compose-hot-reload-gutter-run.png" width="400">
+* For Windows/Linux, select **Run 'MainKt'**.
```
sellmair: This is the wrong task: It should be called MainKt [jvm] for Kotlin Multiplatform projects, it can't be executed from the IDE for regular JVM projects; 

The requirement is not MacOS, but having the Kotlin Multiplatform IDE plugin installed which is now available on Linux and Windows as well.

sarahhaggarty: Ok, I fixed the name and rephrased the different options for running Compose Hot Reload to be about whether the plugin is installed. Does it look better now?

PR #398, opened 2025-10-10
DevTools: Fix DtAnimatedWindow not calling 'tryForceToFront'

sellmair:
The lambda, listening for 'ApplicationWindowGainedFocus' events will be remembered and just captures the value of 'visible' at the time of its creation. This leads to the minimized version of the window always being considered 'visible' whereas the expanded version always reporting 'visible=false'.

We use a custom 'LaunchedEffect' declaring 'title' and 'visible' as keys.


sellmair:
Cherry picked to 1.0.0 ☑️

PR #401, opened 2025-10-21
New OrchestrationMessage encoding

sellmair:
Add OrchestrationMessageEncoder APIs and migrate messages to custom encodings (over j.io.Serializable)

The migration away from j.io.Serializable is supposed to make the protocol
easier to migrate, more predictable, and ready for native & web.

### DIFF
--- a/tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/OrchestrationListenerIntegrationTest.kt
+++ b/tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/OrchestrationListenerIntegrationTest.kt
@@ -9,6 +9,7 @@ import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.flow.toList
 import kotlinx.coroutines.future.await
 import kotlinx.coroutines.job
+import org.jetbrains.compose.reload.core.HotReloadProperty
 import org.jetbrains.compose.reload.core.await
 import org.jetbrains.compose.reload.core.createLogger
 import org.jetbrains.compose.reload.core.getOrThrow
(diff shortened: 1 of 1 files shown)


hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationFrame.kt
```diff
@@ -155,6 +152,7 @@ internal enum class OrchestrationPackageType(val intValue: Int) {
     StateUpdate(3),
     StateUpdateResponse(4),
     StateValue(5),
+    EncodedMessage(6),
```
AbdullinAM: Minor:

The name `EncodedMessage` in this context feels a bit too general and maybe confusing. `JavaSerializableMessage` is technically also an encoded message, but it has its own message type. And it leads to names like `decodeEncodedOrchestrationMessage` and `decodeSerializableOrchestrationMessage` which not the most obvious and readable.

However, calling it something other (e.g. `CustomEncodedMessage`) can also be confusing.

Maybe we should add some comments here explicitly explaining that we need to keep `JavaSerializableMessage` as a separate message type for compatibility reasons

sellmair: Good point! I'll sleep over the names and try to come up with something more intuitive!

sellmair: I updated the code to call the new messages just 'Message'.
Similarly to how other 'custom encoded' types are just called 'StateUpdate' or 'StateUpdateResponse'. 
The special case, here, seems to be the 'JavaSerializableMessage'. So I think just giving it the 'special long name' might make sense. WDYT?

AbdullinAM: Yeah, makes sense 👍

sellmair:
I see one weird test failure with recent changes. 
The incoming commits are to experiment with the issue. 
The review is 'on hold' until I have figured out what is going on!

PR #402, opened 2025-10-21
Recompiler: Automatically use configuration cache and parallel execution for recompilations

sellmair:


properties.yaml
```diff
@@ -145,6 +145,16 @@ GradleWarmupEnabled:
...
     - false: No warmup request will be sent, first reload request may take longer time
 
+GradleBuildOptimize:
+  key: compose.reload.gradle.build.optimize
+  type: boolean
+  default: "true"
+  target: [ application, devtools, build ]
+  isDelicateApi: true
+  documentation: |
+    - true: Compose Hot Reload will try to optimize your build (e.g. by enabling Gradle's configuration cache)
```
pjBooms: > try to optimize your build

try to optimize your _recompilations_? 
Won't it touch regular builds, right?

sellmair: Exactly! This is only attempted during 'hot reload' builds (recompiler)

pjBooms: Then maybe rephrase it to point to it? As now it sounds a bit confusing

sellmair: Done, please check

properties.yaml
```diff
@@ -145,6 +145,17 @@ GradleWarmupEnabled:
...
     - false: No warmup request will be sent, first reload request may take longer time
 
+GradleBuildOptimize:
+  key: compose.reload.gradle.build.optimize
+  type: boolean
+  default: "true"
+  target: [ application, devtools, build ]
+  isDelicateApi: true
+  documentation: |
+    - true: Compose Hot Reload will try to optimize your build, during hot reload
```
pjBooms: I guess comma is not needed here

PR #407, opened 2025-11-03
Simplified DevTools design

AbdullinAM:


properties.yaml
```diff
@@ -210,17 +210,6 @@ DevToolsDetached:
   documentation: |
     If enabled, dev tools window will be detached from the main application
 
-DevToolsAnimationsEnabled:
```
sellmair, line 213: Since this is a binary incompatible change, I think we should go with deprecating it instead

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/devTools.kt
```diff
@@ -86,8 +86,5 @@ private fun platformSpecificJvmArguments(): Array<String> = when (Os.current())
         "-Dapple.awt.UIElement=${!devToolsDetached}",
         "-Dapple.awt.application.name=Compose Hot Reload Dev Tools",
     )
-    // We need no enable OpenGL rendering for skiko, because
```
sellmair, line 89: Nice!

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtAnimatedWindow.kt
```diff
@@ -93,16 +74,14 @@ fun DtAnimatedWindow(
     // If you set `alwaysOnTop = true` for an invisible window in Linux, the window will lose its
     // ability to be always on top (even when it becomes visible again)
     val alwaysOnTopValue = alwaysOnTop && if (Os.currentOrNull() == Os.Linux) visible else true
+    val initialPosition = animateWindowPosition(windowState.position, size, true)
```
sellmair: I noticed being confused upon reading this line for the first time: 
Its called 'initialPosition' which lead me to think of it as 'not changing', 'just an initial value', however the RHS states that its literally anmiated?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtAnimatedWindow.kt
```diff
@@ -208,25 +135,19 @@ internal fun animateWindowPosition(
     }
 }
 
-internal fun DpSize.toDimension(): Dimension = Dimension(width.value.toInt(), height.value.toInt())
+internal val attachedSidecarWindowSize: DpSize = DpSize(
+    width = DtSizes.largeLogoSize + DtPadding.small * 2,
+    height = DtSizes.iconSize * 9 + DtPadding.small * 14,
+)
```
sellmair: I think hardcoding the size here would work (how often do we really add elements to this toolbar, right?), however, its certainly not super elegant. 

AFAIK there is a way (with new versions of compose) to auto-size the window to the contents. This seems really nice!

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtAttachedSidecar.kt
```diff
@@ -5,170 +5,81 @@
...
-import org.jetbrains.compose.reload.core.HotReloadEnvironment.devToolsAnimationsEnabled
+import org.jetbrains.compose.reload.core.HotReloadEnvironment
 import org.jetbrains.compose.reload.core.WindowId
+import org.jetbrains.compose.reload.orchestration.OrchestrationMessage
 
 private val cornerShape = if (devToolsUseTransparency) DtShapes.RoundedCornerShape else DtShapes.SquareCornerShape
-private val useAnimatedTransitions = devToolsUseTransparency && devToolsAnimationsEnabled
 
 @Composable
 fun DtAttachedSidecarWindow(
```
sellmair: I think the wording 'detached' makes sense to differentiate the special case, but I constantly notice that my brain needs to resolve the 'attached' wording when reading the code. 

Suggestion:
We have one 'happy path', 'regular' case (which is currently called) 'attached'. 
We have one 'special case' which is called 'detached'. 

How about just differentiating it by calling the detached code 'detached' and regular code would be unclassified. 

e.g. here: 
`DtSidecarWindow` vs `DtDetachedSidecarWindow`

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtSidecar.kt
```diff
@@ -5,170 +5,81 @@
...
+                tint = animateReloadStatusColor(Color.White).value
             )
+
+            DtSidecarReloadButton()
+            DtSidecarResetButton()
+            DtShowLogsButton()
+            DtSidecarRestartButton()
+            DtSidecarCloseButton()
+
+            DtNotificationsButton()
```
sellmair, line 81: The section reads nicely!

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtSidecar.kt
```diff
@@ -178,59 +89,63 @@ internal fun DtMinimizedSidecarWindowContent(
...
+        tooltip = "Close the application",
+        tag = Tag.ActionButton,
+    ) {
+        DtImage(
+            image = DtImages.Image.CLOSE_ICON,
+            modifier = Modifier.size(DtSizes.iconSize),
+            tint = Color.White,
+        )
+    }
+}
```
sellmair, line 151: Read through this file and I felt really fast here: 
Its easy to understand how the UI looks and behaves here. 
Well done!

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtConsole.kt
```diff
@@ -87,7 +92,7 @@ fun DtConsole(
 
 
 @Composable
-private fun scrollbarStyle(): ScrollbarStyle = LocalScrollbarStyle.current.copy(
+internal fun scrollbarStyle(): ScrollbarStyle = LocalScrollbarStyle.current.copy(
```
sellmair: If it shared, then maybe we could put it into some utilities/widgets source file, so that we could easily re-discover our own utilities?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtLogWindow.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    Window(
+        onCloseRequest = { onClose() },
+        state = rememberWindowState(
+            size = DtSizes.defaultWindowSize,
+            position = WindowPosition.Aligned(Alignment.Center)
+        ),
+        title = "Compose Hot Reload Logs"
+    ) {
+        LaunchedEffect(controller.isOpen.value, controller.focusRequests.value) {
```
sellmair: This line of code shows the creativity of the solution, using your controller, but it also shows its flaws:

This line here, in particular, is a small puzzle: 
The state read of 'isOpen' or 'focusRequests' binds the composable to change to this state; Since the focusRequests will increment they will change, as long as the window is open it will lead to a 'window.toFront' being called.

Let's try out some other approaches to build this logic without the puzzle: 
The idea here is to use more straightforward models

Instead of using the a State<Int> to model the focus requests, we could use a regular flow as something to push a simple 'signal' (Flow<Unit>)
we can then Launch a single coroutine that uses

```kotlin
focusRequests.collectLatest {
    window.toFront()
    window.requestFocus()
}
```

this would model that we want to react to the latest signal instead of going through a state binding, and a keyed LaunchedEffect

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/NotificationsUIState.kt
```diff
@@ -0,0 +1,157 @@
...
+
+fun CoroutineScope.launchNotificationsUIState() = launchState(NotificationsUIState) {
+    fun createNotification(
+        type: UINotificationType,
+        title: String,
+        message: String,
+        isDisposableFromUI: Boolean,
+        details: List<String> = emptyList()
+    ) = UINotificationImpl(type, title, message, isDisposableFromUI, details) { notification ->
+        this@launchNotificationsUIState.launch {
```
sellmair: While this code works and is ergonomic on the side which calls the 'dispose', I think this particular code-site indicates the problem with the design of mixing data and functionality.  

To me, it's easier to expose state to the UI as 'just state'. 
Here, we're conflating it with logic as well. 

Suggestion: Instead of communicating back with this state actor, using the lambda in the state, define an Event here like `NotificationDisposedEvent`: The UI can send such an event, the state actor then handles it and transforms the state.

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/NotificationsUIState.kt
```diff
@@ -0,0 +1,157 @@
...
+                    details = state.logs.map { it.message },
+                )
+                NotificationsUIState.update { it + currentStateNotification }
+            } else {
+                currentStateNotification?.dispose()
+            }
+        }
+    }
+
+    checkForJBR()
```
sellmair: I wonder why you have decided to 'block' the coroutine here for checking the JBR?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/NotificationsUIState.kt
```diff
@@ -0,0 +1,157 @@
...
+    val type: UINotificationType
+    val title: String
+    val message: String
+    val isDisposableFromUI: Boolean
+    val details: List<String>
+
+    fun dispose()
+}
+
+private class UINotificationImpl(
```
sellmair: If you follow my suggestion of separating logic (disopose) from the data, then I think we could also remove the interface for a simple data class

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtTooltip.kt
```diff
@@ -0,0 +1,136 @@
...
+import org.jetbrains.compose.devtools.theme.DtTextStyles
+import kotlin.time.Duration.Companion.milliseconds
+
+private val tooltipShowDelay = 500.milliseconds
+private val tooltipHideDelay = 100.milliseconds
+
+private val defaultTooltipOffset = DpSize(1.dp, 1.dp)
+
+@Composable
+fun DtTooltip(
```
sellmair, line 50: Wow this is a really cool API!

hot-reload-devtools/src/test/kotlin/org/jetbrains/compose/devtools/tests/state/NotificationsUIStateTest.kt
```diff
@@ -0,0 +1,165 @@
...
+import org.jetbrains.compose.reload.core.complete
+import org.jetbrains.compose.reload.core.reloadMainThread
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+
+class NotificationsUIStateTest {
```
sellmair, line 37: 😍

hot-reload-gradle-plugin/api/hot-reload-gradle-plugin.api
```diff
@@ -12,7 +12,6 @@ public abstract interface class org/jetbrains/compose/reload/gradle/ComposeHotRe
 	public abstract fun isRecompilerWarmupEnabled (Lorg/gradle/api/provider/Provider;)V
 	public abstract fun setAgentJar (Lorg/gradle/api/file/FileCollection;)V
 	public abstract fun setArgFile (Lorg/gradle/api/provider/Provider;)V
-	public abstract fun setDevToolsAnimationsEnabled (Lorg/gradle/api/provider/Provider;)V
```
sellmair, line 15: Let's just deprecate that!

sellmair (changes requested):
Really good job!!! Can't wait to ship 1.1!

properties.yaml
AbdullinAM: Yeah, makes sense. I was too eager to remove the unnecessary property and forgot about compatibility 😅

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtAnimatedWindow.kt
AbdullinAM: Yeah, it is indeed confusing. The RHS is not animated here, as we pass last argument `skipAnimation = true`. This is just an artifact of the past that I forgot to change, will update

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/states/NotificationsUIState.kt
AbdullinAM: I think this was just forgotten, I'll update

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/sidecar/DtAnimatedWindow.kt
AbdullinAM: I currently only know of the way to auto-size the window to fit its contents on creation by setting its size to `Unspecified`. But our window changes its height dynamically, so this solution does not work.

If there are ways to dynamically adjust the window size --- that would be really cool! I will look into it

AbdullinAM: https://youtrack.jetbrains.com/issue/CMP-7377/JVM-Resize-Window-size-to-fit-content

There are workarounds, but people report that they cause flicker

sellmair: Then lets just contiue like originally proposed by you!

PR #412, opened 2025-11-07
Remove JVM hijack tasks

### LINKED ISSUE #96: Warning: Target 'jvmDesktop': Unable to create run task 'jvmDesktopRun' as there is already such a task registered
If I add compose-hot-reload to project. when I sync gradle, a warning appear.

```
Target 'jvmDesktop': Unable to create run task 'jvmDesktopRun' as there is already such a task registered
Please remove the conflicting task or rename the new task
```

AbdullinAM:
Fix #96: we don't need JVM hijack tasks anymore, as KMP plugin is available on all OS's and has a proper IDE support for hot reload


sellmair:
Cherry-picked into 1.0.0 ☑️

PR #422, opened 2025-11-14
Remove the custom blank issue template and explicitly enable the built-in one

AbdullinAM:
After recent changes in the GH issue templates, our 'Create an issue' chooser has two 'Blank issue' options: one is default and the other is our custom.

Screenshot 2025-11-14 at 10 53 38

New there will be only one


AbdullinAM:
@sellmair I think we can skip the CI requirements for this one, as it is only changing the GH infrastructure

PR #427, opened 2025-11-17
[runtime] Initialize `WindowState` only on receiving window events

AbdullinAM:
This ensures that the devtools window receives the information about the app window only after it has been created and made visible

### DIFF
--- a/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
+++ b/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
@@ -69,7 +69,9 @@ internal fun startWindowManager(window: Window): WindowId {
             OrchestrationMessage.ApplicationWindowGone(windowId).sendAsync()
         }
 
-        broadcastActiveState()
+        if (window.isActive) {
+            broadcastActiveState()
+        }
 
         val windowListener = object : WindowAdapter() {
             override fun windowIconified(e: WindowEvent?) {
@@ -101,6 +103,12 @@ internal fun startWindowManager(window: Window): WindowId {
         }
 
         val componentListener = object : ComponentAdapter() {
+            fun broadcastIfActive() {
+                if (window.isActive) {
+                    broadcastActiveState()
+                }
+            }
+
             override fun componentHidden(e: ComponentEvent?) {
                 logger.trace { "$windowId: componentHidden" }
                 broadcastGone()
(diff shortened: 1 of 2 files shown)


hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -51,7 +51,9 @@ internal fun startWindowManager(window: Window): WindowId {
             }
         }
 
-        fun broadcastActiveState() {
+        fun broadcastActiveState(isActive: Boolean) {
+            if (!isActive) return
```
sellmair: I think this is a little confusing: 
As a reader, I wonder why you would not inline `window.isActive`, 
however, it seems that you just guard the component listener with `window.isActive`. 
I would therefore prfer the `componentListener` to do the guard itself. 
Btw, is this listener actually necessary? Maybe just removing it does the trick as well?

AbdullinAM: Yeah, moving the check to the component listener makes sense 👍 

> Btw, is this listener actually necessary? Maybe just removing it does the trick as well?

I'll check

AbdullinAM: * Moved the guard to the caller side
* `componentListener` is needed, because it allows us to react to window movement/resizing. Regular `windowListener` only has callbacks for activation/deactivation of the window

tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/WindowIntegrationTest.kt
```diff
@@ -109,9 +117,46 @@ class WindowIntegrationTest {
             """.trimIndent()
     )
 
-    private fun runDevelopmentEntryPointTest(
+    @Headless(false)
+    @ReloadEffects
+    @HostIntegrationTest
+    @HotReloadTest
+    fun `test - Window with late activation`(
```
sellmair, line 133: I am OK with this, but I have some concerns about how this is abstracted to share code. 
The test code here now requires some additional parsing and understanding of the abstraction instead of delegating some work to utility functions. 

Again, I do not want to nudge you here, but I wanted to point out, that I feel like plainly duplicating some code in tests seems easier?

sellmair (approved):

PR #434, opened 2025-11-19
[properties] Add `visibility` field to parameters

AbdullinAM:
* Remove `@InternalHotReloadApi` from `HotReloadEnvironment` and `HotReloadProperty`, mark every individual property with a proper visibility annotation
* Add a separate `visibility` field in `properties.yaml`. Visibility has 5 possible values: public, delicate, experimental, internal, deprecated. Internal by default. You can add `deprecationMessage` when deprecating properties
* All generated code related to a property is marked with a corresponding annotation
* I decided to now change visibility of `ComposeHotReloadArgumentsBuilder`. In general it is not ideal that we have to manage properties in two separate places, but I don't know what is the best way to change that. At least now we will get deprecation warnings if we decide to deprecate one of the properties
* Add an explicit `isUserConfigurable` field to properties to differentiate the properties set by CHR from properties expected from the user. We need this field, because some of the properties are public, but not expected from the user (e.g. `IsHotReloadActive`)

Potential next step: 
We can use `HotReloadProperty` API on the IDE side to automatically generate UI for hot run configuration options. E.g. we can generate a field for each property with `visibility == Visibility.Public` and `isUserConfigurable == true`

### DIFF
--- a/properties.yaml
+++ b/properties.yaml
@@ -127,7 +127,7 @@ GradleBuildContinuous:
   type: boolean
   default: "false"
   target: [ application, devtools, build ]
-  isDelicateApi: true
+  visibility: delicate
   documentation: |
     - true: Compose Hot Reload will start a recompiler Gradle Daemon, which will continuously rebuilt/reload the project
     by watching all  inputs to the build
@@ -139,7 +139,7 @@ GradleWarmupEnabled:
   type: boolean
   default: "false"
   target: [ application, devtools, build ]
-  isDelicateApi: true
+  visibility: delicate
   documentation: |
     - true: Compose Hot Reload will launch a warmup recompile request when the application is started, to ensure that
     the recompiler Gradle daemon is running and ready to handle requests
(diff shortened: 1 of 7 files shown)


properties.yaml
```diff
@@ -308,7 +321,8 @@ StdinFile:
   key: compose.reload.stdinFile
   type: file
   target: [ build, application, devtools ]
-  isDelicateApi: true
+  visibility: delicate
+  isUserConfigurable: true
```
sellmair: QQ: Do you have use cases in mind?

AbdullinAM: I had two main use cases:
* Potential automatic export of properties to IDE
* Semi-documentation: we don't have any docs about our properties, and just reading through `properties.yaml` it is not always clear which properties are provided by CHR and which ones are expected from the user

However, thinking about it more and more, I think that the idea of automatic import of properties to IDE is not really feasible. Because IDE will also want to properly bundle and localize all the names/descriptions of the properties.

So I'm still doubting if I want to keep `isUserConfigurable`, as I don't really like how it looks

sellmair: I would not have cared for such import about localization; We're importing an API that was written in English and exposing those functions in English. I, however, won't push on this feature

PR #437, opened 2025-11-25
[devtools] Explicitly receive local events to emit UINotificationDisposeEvent

AbdullinAM:
`rememberCoroutineScope` is not guaranteed to inherit the application coroutine context, so we have to ensure we get it in the composable scope


hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtNotificationCard.kt
```diff
@@ -86,7 +87,9 @@ fun DtNotificationCard(notification: UINotification) {
                 if (notification.isDisposableFromUI) {
                     DtIconButton(
                         onClick = {
-                            scope.launch { UINotificationDisposeEvent(notification.id).emit() }
+                            scope.launch {
```
sellmair: I think using `EvasLaunching {}` is the simpler approach

PR #440, opened 2025-11-28
Use isolated project access API

simonlebras:
Using the isolated project access API is required for compatibility with Gradle's configuration caching.

### DIFF
--- a/hot-reload-core/api/hot-reload-core.api
+++ b/hot-reload-core/api/hot-reload-core.api
@@ -240,6 +240,7 @@ public final class org/jetbrains/compose/reload/core/HotReloadEnvironment {
 	public final fun getGradleBuildContinuous ()Z
 	public final fun getGradleBuildOptimize ()Z
 	public final fun getGradleWarmupEnabled ()Z
+	public final fun getIsolatedProjectsEnabled ()Z
 	public final fun getLogLevel ()Lorg/jetbrains/compose/reload/core/Logger$Level;
 	public final fun getLogStdout ()Z
 	public final fun getReloadEffectsEnabled ()Z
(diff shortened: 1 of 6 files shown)


AbdullinAM:
Hi! Thank you for opening a PR!

Unfortunately, I cannot approve this one. Isolated projects are a [pre-alpha](https://docs.gradle.org/current/userguide/isolated_projects.html) Gradle feature, that only **extends and improves** configuration cache and it is not required for it. We can't afford to introduce a dependency on an incubating feature that changes from one version of Gradle to other. You may notice, that even our tests start failing with `Caused by: java.lang.NoSuchMethodError: 'org.gradle.api.project.IsolatedProject org.gradle.api.Project.getIsolated()'` on Gradle 8.7 with this change.


In Compose Hot Reload we aim to provide compatibility with a wide range of versions for all the tools/libraries users may use. Therefore we try to rely on only stable APIs

simonlebras:
Hi !

Thank you for your feedback.

Would it be better to guard this like KSP did with a custom [Gradle property](https://github.com/google/ksp/pull/2713/files#diff-391494f1af781fc356039fae284af609f0c67281e2b626fa22a26222f82d695dR26) ?
or even with that, it is still to dangerous ?

AbdullinAM:
I think that a custom flag can be a good solution, yes! Good idea!

simonlebras:
Fixed with [12b8c7b](https://github.com/JetBrains/compose-hot-reload/pull/440/commits/12b8c7b143bb0a716beb465b3d2358a000a72c5b)

sellmair:
Not required (we can do this for you), but I think writing a functional test, checking if isolated projects work, would be fine. 
There is a similar smoke test that you could look at for inspiration:

https://github.com/JetBrains/compose-hot-reload/blob/706b5dc5fdfffb0c9d4dc170769d2f5c974142d0/tests/src/reloadFunctionalTest/kotlin/org/jetbrains/compose/reload/tests/gradle/TasksSmokeTest.kt#L21

But I think, once @AbdullinAM accepts the contribution, you'd be fine. Writing the test should not be required; we can do this for you.

PR #441, opened 2025-11-28
Fix DevTools window not appearing, if we have an error in the composition.

### LINKED ISSUE #322: App fails when restart after exception
If some exception is thrown and shown in hot reload mod and you click to "Reload" then application is restarted and failed, gradle task is stopped.

https://github.com/user-attachments/assets/db49a861-cfd8-4641-b573-3fc0c4fe8942

pjBooms:
Fixes #322

If we have an error in the composition we may not launch LaunchedEffects. 
Previously we had a logic for window management (synchronizing the app window with the devtool window) in a LaunchedEffect that could be skipped in case of errors, thus devtools was not notified about app window appearing and did not show devtools window.

Here we fix it launching the required logic instantly.

### DIFF
--- a/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
+++ b/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
@@ -6,15 +6,12 @@
 package org.jetbrains.compose.reload.jvm
 
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.RememberObserver
 import androidx.compose.runtime.remember
-import kotlinx.coroutines.awaitCancellation
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.trySendBlocking
-import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.flow.conflate
 import kotlinx.coroutines.flow.consumeAsFlow
-import kotlinx.coroutines.job
 import org.jetbrains.compose.devtools.api.WindowsState
 import org.jetbrains.compose.devtools.api.WindowsState.WindowState
 import org.jetbrains.compose.reload.agent.orchestration
@@ -37,7 +34,7 @@ private val logger = createLogger()
 internal fun startWindowManager(window: Window): WindowId {
     val windowId = remember { WindowId.create() }
 
-    LaunchedEffect(windowId) {
+    remember(windowId) {
         val windowState = Channel<WindowState?>(Channel.CONFLATED)
 
         /* Synchronize windows state with orchestration */
@@ -139,15 +136,17 @@ internal fun startWindowManager(window: Window): WindowId {
         window.addWindowFocusListener(windowListener)
         window.addComponentListener(componentListener)
 
-        currentCoroutineContext().job.invokeOnCompletion {
-            window.removeWindowListener(windowListener)
-            window.removeComponentListener(componentListener)
-            broadcastGone()
-            windowState.close()
+        object : RememberObserver {
+            override fun onRemembered() {}
+            override fun onAbandoned() {}
+
+            override fun onForgotten() {
+                window.removeWindowListener(windowListener)
+                window.removeComponentListener(componentListener)
+                broadcastGone()
+                windowState.close()
+            }
         }
-
-        awaitCancellation()
     }
-
     return windowId
 }


hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -33,121 +27,124 @@ import java.awt.event.WindowEvent
...
-
-        /* Synchronize windows state with orchestration */
-        orchestration.subtask {
-            windowState.consumeAsFlow().conflate().collect { state ->
-                orchestration.update(WindowsState) { current ->
-                    val windows = if (state == null) current.windows - windowId
-                    else current.windows + (windowId to state)
-                    WindowsState(windows)
-                }
+    curWindowId?.let {
```
AbdullinAM: Although this approach works good for one window, I think it will break if a single application launches multiple windows

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -33,121 +27,124 @@ import java.awt.event.WindowEvent
 
 private val logger = createLogger()
 
-@Composable
+private var curWindowId: WindowId? = null
+private var curWindow: Window? = null
```
sellmair: There can be multiple windows, so I think this code is violating that assumption. 
As a second note: I think holding a strong reference to an object like window, statically might be a bad idea.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -33,121 +27,124 @@ import java.awt.event.WindowEvent
...
+    }
+
+    val windowId = WindowId.create()
+    curWindowId = windowId
+    curWindow = window
+
+    val windowState = Channel<WindowState?>(Channel.CONFLATED)
+
+    /* Synchronize windows state with orchestration */
+    orchestration.subtask {
```
sellmair: This now will launch one subtask for each composition, which can be a lot.

AbdullinAM: I have a concern that this `subtask` will never finish execution.

Previously we had a clean-up procedure: removed listeners (I'm not sure if its necessary) and closed the channel, which lead to the `subtask` completion. Now the state is never closed and the subtask is never finished. 

Realistically, 99% of the applications have only one window and stop the process when it is closed. Can we at least test this scenario and see if there is a 'leak' and how bad it is?

AbdullinAM (changes requested):

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: The function is called only once in the fix, so it is not the case.

sellmair: Error was captured in composition while live edit was enabled.
java.lang.IllegalStateException: unexpected window re-initialization

image

The function is called within a `@Composable` function, so it can be called as often as with every frame (roughly).

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: Rewrote. Please take a look.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: But we returned immediately from `startWindowManager`  not reaching this point (` orchestration.subtask` was called only once for a window). Anyway, I rewrote it.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: addressed

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: Addressed. Now we do the cleanup  on the composition exit.

AbdullinAM (approved):
Nice! I like how the fixed version looks 🔥

PR #444, opened 2025-12-03
Update README.md

cedricferry:
by default KMP plugin creates a project with "composeApp" name, using "app" in the documentation can be misleading. added an example with `composeApp`. other section may require update too.


sellmair:
I do like the contribution, thank you!

PR #446, opened 2025-12-03
Cleanup tests

AbdullinAM:


hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ignore.kt
```diff
@@ -61,6 +61,7 @@ interface Ignore {
             "kotlin/", "kotlinx/", "androidx/",
             "org/jetbrains/skia/", "org/jetbrains/skiko/",
             "org/jetbrains/compose/reload",
+            "org/jetbrains/compose/hotReloadUI",
```
sellmair, line 64: Wait, the entire reason for choosing this strange package was to allow for using hot reload when working on the UI, adding it to the ignore kind of defeats the purpose?

AbdullinAM: Makes sense. When experimenting with the source information markers I noticed that the ui package started leaking into the analysis. That got me confused, it does not happen on the master version.

I'll revert the change. Thank you for catching!

PR #451, opened 2025-12-05
Tests cleanup

AbdullinAM:
~~-[tests] Apply `@Execution(ExecutionMode.SAME_THREAD)` for `GradleRecompilerProcessTest.test - gradle recompiler process is stopped - ShutdownRequest`~~

### DIFF
--- a/hot-reload-test/gradle-testFixtures/src/main/kotlin/org/jetbrains/compose/reload/test/gradle/HotReloadTestFixtureExtension.kt
+++ b/hot-reload-test/gradle-testFixtures/src/main/kotlin/org/jetbrains/compose/reload/test/gradle/HotReloadTestFixtureExtension.kt
@@ -22,6 +22,7 @@ import java.nio.file.Files
 import kotlin.io.path.ExperimentalPathApi
 import kotlin.io.path.exists
 import kotlin.io.path.listDirectoryEntries
+import kotlin.io.path.name
 
 internal class HotReloadTestFixtureExtension(
     private val context: HotReloadTestInvocationContext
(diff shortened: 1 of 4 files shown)


sellmair:
Hmm, we do not run tests in parallel on CI still IIRC, therefore I assume that such an annotation should not have any effect on how it gets executed or its flakiness.

AbdullinAM:
Yes, this was only one of my random theories, and it did not prove to be correct. 
Overall I'm still struggling to reproduce the issue locally. I have a theory on what is happening, but don't understand how it can happen.

I'll keep experimenting🙂

AbdullinAM:
Although, I think I found a couple of other small issues in tests and I'm currently fixing them

hot-reload-test/gradle-testFixtures/src/main/kotlin/org/jetbrains/compose/reload/test/gradle/HotReloadTestFixtureExtension.kt
```diff
@@ -107,6 +108,9 @@ internal class HotReloadTestFixtureExtension(
             extras[CheckScreenshot.key] = checkScreenshot
         }
 
+        // Save the shutdown log of the application
+        extras[SaveExecutionLogs.key] = SaveExecutionLogs(hotReloadLogsDirectory) { it.name == "shutdown.log" }
+
```
sellmair, line 113: If the extra already lives as implementation inside the fixture, then we could make it much simpler just by inlining this 'shutdown.log' into 'saveExecutionLogs'  and get rid of the indirection?

AbdullinAM: Makes sense, yes 👍

PR #453, opened 2025-12-08
Simplify support for Gradle isolated projects

simonlebras:
Support for isolated projects should be simpler that what I introduced with #440. [rootDir](https://docs.gradle.org/current/userguide/isolated_projects.html#build_logic_constraints) is already compatible with isolated projects, so need for a custom property.

### DIFF
--- a/properties.yaml
+++ b/properties.yaml
@@ -382,3 +382,12 @@ IsHotReloadActive:
   documentation: |
     Will be set to 'true' if the application is launched with Hot Reload and therefore can be used
     to detect if hot reload is 'active'
+
+IsolatedProjectsEnabled:
+  key: compose.reload.isolatedProjectsEnabled
+  type: boolean
+  target: [ application, build, devtools ]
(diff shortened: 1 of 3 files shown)


properties.yaml
```diff
@@ -382,13 +382,3 @@ IsHotReloadActive:
   documentation: |
     Will be set to 'true' if the application is launched with Hot Reload and therefore can be used
     to detect if hot reload is 'active'
-
-IsolatedProjectsEnabled:
```
AbdullinAM, line 386: I would suggest not deleting the property right away, but deprecating it for at least one alpha release cycle. Just to be safe. But again, we can do this ourselves after merge

simonlebras: I agree. 👍 Fixed with [96ed8e3](https://github.com/JetBrains/compose-hot-reload/pull/453/commits/96ed8e3c9af20950a61625e0175322054f46500d)

PR #454, opened 2025-12-09
Recompiler: Enable IC for KMP

sellmair:
The property has a significant impact on build times. There are some known 'unsafe' issues, e.g., the following code.

```
// common
// FILE: A.kt
fun foo(x: Any) = 10
// FILE: B.kt
fun main() {
   println(foo(1))
}
// platform
// FILE: C.kt
fun foo(x: Int) = 20
```

Would print either 10 or 20, depending on which of A.kt and C.kt is clean.

Such cases, while not ideal for 'real builds,' are not necessarily deal-breakers for trusting hot reload as a tool.


sellmair:
It's not worth trying to explain the situation. It's either intuitively _good enough_ or we should make it opt in instead.

PR #455, opened 2025-12-09
Support TableSwitch and LookupSwitch instructions in the InstructionTree

AbdullinAM:


AbdullinAM:
Results of the instruction tree parse benchmarks: switch statements did not worsen the overall parsing speed

Screenshot 2025-12-10 at 10 57 13

Issue #230, opened 2025-06-17
Top level vals are not re-initialized if their initializing methods are in another class and changed


pjBooms:
Consider we have a Kotlin file:

`Val.kt:`
```kotlin
val greeting = getGreeting()
```

where `getGreeting()` in another file:

`Greeting.kt:`
```kotlin
fun getGreeting() = "Hello"
```

If we change `getGreeting()` body `val greeting` will remain the same.

https://github.com/user-attachments/assets/4ff7c0db-66f9-45c5-a5be-e0e3329da19f

sellmair:
Oh oh! Sounds like we do not properly use the dirty resolution here!

sellmair:
Oh wait, I see the code, it was done somewhat on purpose! Let's connect in person to see check if my original line of thoughts are still OK or can be changed freely!

PR #456, opened 2025-12-09
Introduce AllDirty mode for statics (top level vals) re-initialization

pjBooms:
When "AllDirty" is set, all static initializers that are recognized as dirty (affected by changes) will be re-invoked on reload. Previously, only statics of changed classes were re-initialized ("ChangedOnly" mode).

Fixes: #230

Test: test added to statics.kt of reloadUnitTest. KotlinConfApp is manually tested

### DIFF
--- a/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
+++ b/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
@@ -7,12 +7,19 @@ package org.jetbrains.compose.reload.agent
 
 import javassist.CtClass
 import javassist.CtConstructor
+import org.jetbrains.compose.reload.analysis.MethodId
 import org.jetbrains.compose.reload.analysis.classId
 import org.jetbrains.compose.reload.analysis.classInitializerMethodId
+import org.jetbrains.compose.reload.analysis.isClassInitializer
+import org.jetbrains.compose.reload.core.HotReloadEnvironment
+import org.jetbrains.compose.reload.core.StaticsReinitializeMode
 import org.jetbrains.compose.reload.core.createLogger
 import org.jetbrains.compose.reload.core.debug
+import org.jetbrains.compose.reload.core.error
 import org.jetbrains.compose.reload.core.sortedByTopology
 import org.jetbrains.compose.reload.core.warn
+import java.lang.instrument.ClassDefinition
+import java.lang.instrument.Instrumentation
 import java.lang.reflect.Modifier
 
 private val logger = createLogger()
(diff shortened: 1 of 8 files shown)


hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
```diff
@@ -51,24 +57,57 @@ internal fun CtClass.transformForStaticsInitialization(originalClass: Class<*>?)
...
+    // Transform classes for re-initialization that are not transformed yet
+    val toRedefine = dirtyClasses.mapNotNull { clazz ->
+        if (runCatching { clazz.getDeclaredMethod(reinitializeName) }.isSuccess) return@mapNotNull null
+
+        runCatching {
+            val inputStream = clazz.classLoader.getResourceAsStream(clazz.classId.value + ".class")
+            val transformedClass = getClassPool(clazz.classLoader).makeClass(inputStream)
+            transformedClass.transformForStaticsInitialization(clazz)
+            ClassDefinition(clazz, transformedClass.toBytecode())
+        }.getOrNull()
```
sellmair: Maybe we should log the such issue on 'warning' or 'debug' level?

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
```diff
@@ -51,24 +57,57 @@ internal fun CtClass.transformForStaticsInitialization(originalClass: Class<*>?)
...
+        }
+        StaticsReinitializeMode.AllDirty -> {
+            reload.dirty.dirtyMethodIds.keys.filter { it.isClassInitializer }
+                .map(MethodId::classId).mapNotNull { classId ->
+                    val loader = findClassLoader(classId).get()
+                    if (loader == null) {
+                        logger.debug("Class '$classId' is not loaded yet")
+                        return@mapNotNull null
+                    }
+                    return@mapNotNull runCatching { loader.loadClass(classId.toFqn()) }.getOrNull()
```
sellmair: Maybe logging the issue _could_ be helpful later?

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
```diff
@@ -51,24 +57,57 @@ internal fun CtClass.transformForStaticsInitialization(originalClass: Class<*>?)
...
+
+        runCatching {
+            val inputStream = clazz.classLoader.getResourceAsStream(clazz.classId.value + ".class")
+            val transformedClass = getClassPool(clazz.classLoader).makeClass(inputStream)
+            transformedClass.transformForStaticsInitialization(clazz)
+            ClassDefinition(clazz, transformedClass.toBytecode())
+        }.getOrNull()
+    }
+    if (toRedefine.isNotEmpty()) {
+        check(HotReloadEnvironment.staticsReinitializeMode != StaticsReinitializeMode.ChangedOnly) {
```
sellmair: Good assertion. Let's think about if it should just be a warning instead of aborting the reload. Statics re-init is a 'best effort' not necessarily a dealbreaker?

tests/src/reloadUnitTest/kotlin/tests/statics.kt
```diff
@@ -35,3 +36,33 @@ fun `test - change top level property`() {
...
+        package tests
+        object StaticsObject {
+            val property = "bar"
+        }
+    """.trimIndent()
+    )
+}
+
+private fun setStaticsReinitializeMode(mode: StaticsReinitializeMode) {
+    System.setProperty("compose.reload.staticsReinitializeMode", mode.name)
```
sellmair: Let's use the HotReloadProperty directly here :)

tests/src/reloadUnitTest/kotlin/tests/statics.kt
```diff
@@ -35,3 +36,33 @@ fun `test - change top level property`() {
...
+}
+
+@HotReloadUnitTest
+fun `test - change initializer of top level property (ChangedOnly mode)`() {
```
sellmair, line 58: Good tests 🤙

properties.yaml
```diff
@@ -392,3 +392,15 @@ IsolatedProjectsEnabled:
   documentation: |
     Enables support for Gradle's incubating 'isolated projects' feature
 
+StaticsReinitializeMode:
+  key: compose.reload.staticsReinitializeMode
+  type: enum
+  enumClass: org.jetbrains.compose.reload.core.StaticsReinitializeMode
+  target: [ application ]
+  default: "AllDirty"
```
sellmair: I would advise keeping the original behavior as default first and collecting user projects or code snippets which benefit from the changes, as well as projects and code snippets which would fail because of the 'AllDirty' mode.

On the one hand, having more up-to-date statics is desirable, on the other hand, re-initializing all statics (including final ones, including properties which technically are not dirty) could have severe unwanted side effects as well. I cannot predict the weight of both effects, therefore I suggest that 'real life applicability research' shall be the next step.

AbdullinAM (commented):
Good job! I like the solution 🔥 

I overall agree with Seb's comments

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
pjBooms: Done

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
pjBooms: Logged as error to potentially see the exeception

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/staticsInitialization.kt
pjBooms: Agree. As we don't fail on failed attempt to transform earlier, it sounds ok to just warn here

tests/src/reloadUnitTest/kotlin/tests/statics.kt
pjBooms: Done

properties.yaml
pjBooms: I agree with 'real life applicability research'. We just need to agree with the strategy. 

I set "AllDirty" hoping to gather instant feedback with the next alpha as alphas is good moment to check experimental things. My intent was: if we would have problems with the mode, we'd get them earlier in case we enable the mode by default, so we could decide to return the behavior to original or not. 

However I am ok with gradual adoption strategy as well: keep the original behavior with the next alpha, asking users to enable the mode and provide the feedback. If no problems with it will be reported, eventually enable it by default (I guess we should do it with some alpha anyway).

My intuition says that for our use case - changes in UI code - "AllDirty" mode is preferable:

1. I hardly expect non-trivial code in static initializers in UI-related sources with side effects making them non re-callable. 
2. If it would happen to have such static initializers, I believe there are almost equal chances to explode on them for both modes. Really, if we deduce that such an initializer is affected by a change, it means that the problematic code is really-really close to the change and the chances are high that the next change will be just in it. 

But let's see if my intuition is correct.

PR #457, opened 2025-12-16
[runtime] memory optimisations

AbdullinAM:
### Main changes

* Custom string interning based on a weak hash map
* Collection optimisations
  * Use singleton objects whenever possible
  * Use hash maps/sets instead of their linked analogs
  * Replace dependency sets in `ScopeInfo` with lists: we don't need/use their set properties

### Side effects

* Explicitly sort scopes in `Iterable<ScopeInfo>.invalidationKey()` before computing hash. This way we are not relying on the original collections implementation details. Has negligible impact on the `resolveDirty` analysis: ~1ms slower. Collections in `invalidationKey` usually have only 1-2 elements
* Majority of the execution time (~70-80%) in `resolveDirty` is taken by `redefinition.dirtyMethodIds.entries.sortedBy { it.key.methodDescriptor }` and logging. I updated the implementation to only invoke sorting when the logging is enabled

### Results

Changes allow to save ~45-50% memory, especially on big projects. There are still many small memory optimisations that we can implement, but they will only provide small gains at the cost of a bigger impact on efficiency/public API behaviour

Screenshot 2025-12-16 at 12 54 55
Screenshot 2025-12-16 at 12 55 01

### DIFF
--- a/hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
+++ b/hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
@@ -42,16 +42,16 @@ fun ClassInfo(classNode: ClassNode): ClassInfo? {
                 else -> MethodInfo.Modality.OPEN
             }
         )
-    }.associateBy { it.methodId }
+    }.associateByTo(hashMapOf()) { it.methodId }.toReadOnlyHashMap()
 
-    val fields = classNode.fields.associate { fieldNode ->
+    val fields = classNode.fields.associateTo(hashMapOf()) { fieldNode ->
         FieldId(classNode, fieldNode) to FieldInfo(
             fieldId = FieldId(classNode, fieldNode),
             isStatic = fieldNode.access and (Opcodes.ACC_STATIC) != 0,
             initialValue = fieldNode.value,
             additionalChangeIndicatorHash = getResourceContentHash(fieldNode)
         )
-    }
+    }.toReadOnlyHashMap()
 
     return ClassInfo(
         classId = classId,
--- a/hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/Interner.kt
+++ b/hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/Interner.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2024-2025 JetBrains s.r.o. and Compose Hot Reload contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ */
+
+package org.jetbrains.compose.reload.analysis
+
+import java.lang.ref.WeakReference
+import java.util.WeakHashMap
+import java.util.concurrent.locks.ReentrantReadWriteLock
+import kotlin.collections.set
+import kotlin.concurrent.read
+import kotlin.concurrent.write
+
+private class Interner<T> {
+    private val lock = ReentrantReadWriteLock()
+    private val map = WeakHashMap<T, WeakReference<T>>()
+
+    fun intern(value: T): T {
+        val existing = lock.read { map[value]?.get() }
+        if (existing != null) return existing
+
+        return lock.write {
+            /* Check that the value was not added while we waited for the write lock */
+            val added = map[value]?.get()
+            if (added != null) return@write added
+
+            map[value] = WeakReference(value)
+            value
+        }
+    }
+}
+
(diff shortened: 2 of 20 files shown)


hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/asmUtils.kt
```diff
@@ -74,38 +74,38 @@ fun ClassId(bytecode: ByteArray): ClassId? {
...
+    return className?.let { name -> ClassId(name.interned()) }
 }
 
 
 internal fun MethodId(classNode: ClassNode, methodNode: MethodNode): MethodId = MethodId(
     classId = ClassId(classNode),
-    methodName = methodNode.name,
-    methodDescriptor = methodNode.desc,
+    methodName = methodNode.name.interned(),
+    methodDescriptor = methodNode.desc.interned(),
```
sellmair, line 84: Have you checked if interning `MethodId` would also make sense?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
```diff
@@ -42,16 +42,16 @@ fun ClassInfo(classNode: ClassNode): ClassInfo? {
                 else -> MethodInfo.Modality.OPEN
             }
         )
-    }.associateBy { it.methodId }
+    }.associateByTo(hashMapOf()) { it.methodId }.toImmutableHashMap()
```
sellmair: Are you concerned about someone casting it back to MutableMap?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/utils.kt
```diff
@@ -36,3 +36,11 @@ internal fun Checksum.updateString(value: String) {
     updateInt(value.length)
     update(value.encodeToByteArray())
 }
+
+internal fun <K, V> HashMap<K, V>.toImmutableHashMap(): Map<K, V> = when (size) {
+    0 -> emptyMap()
+    1 -> mapOf(entries.first().toPair())
```
sellmair: Maybe using `singletonMap` directly here would make it clearer? This way, a reader needs to know that mapOf with this single pair is optimizing by creating a Singleton Map

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/utils.kt
```diff
@@ -36,3 +36,11 @@ internal fun Checksum.updateString(value: String) {
     updateInt(value.length)
     update(value.encodeToByteArray())
 }
+
+internal fun <K, V> HashMap<K, V>.toImmutableHashMap(): Map<K, V> = when (size) {
```
sellmair: The name seems a little bit off 🤔 
It's not the primary purpose to be immutable (it returns a read-only view that can be mutated if we really want to). How about `toHashMap()` or `toReadOnlyHashMap()`

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/Interner.kt
```diff
@@ -0,0 +1,29 @@
...
+package org.jetbrains.compose.reload.analysis
+
+import java.lang.ref.WeakReference
+import java.util.WeakHashMap
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.collections.set
+import kotlin.concurrent.withLock
+
+private class Interner<T> {
+    private val lock = ReentrantLock()
```
sellmair: Have you tried using a read-write lock?

sellmair (commented):
Looks really good!
I do have two ideas: 

1) What about interning things like `MethodId` as well? Potentially, the intern could just be generally allowing any object to be interned?

2) Have you run benchmarks to check any potential impact?

hot-reload-analysis/src/main/kotlin/org/jetbrains/compose/reload/analysis/ClassInfo.kt
AbdullinAM: The purpose of this change is to:
1. Ensure that we use hash maps for storing methods and fields --- more memory efficient than linked hash map
2. Ensure that we use singleton objects when possible via the call to `toImmutableHashMap`
3. Prevent unnecessary copying, so that we can re-use the result of `associateBy` if the map actually contains elements

sellmair:
Fantastic work 🤙

PR #466, opened 2026-01-22
Update dependency versions

AbdullinAM:


gradle.properties
```diff
@@ -23,3 +23,9 @@ android.useAndroidX=true
 ############
 kotlin.native.enableKlibsCrossCompilation=true
 ############
+# Compose Resources
+############
+# Disable resource content hash generation because some modules use older ([composeMin]) version
+# of Compose Resources that does not support it yet.
+org.jetbrains.compose.resources.content.hash.generation.disable=true
```
sellmair, line 30: If those modules are older, wouldn't they then not care about this flag anyways?

AbdullinAM: Basically the main problem is the `hot-reload-test` module. If we use compose plugin 1.10.0, it generates resource hash annotation for the resources used in this module (font used for the `TestText`). However, then we have problems compiling `hot-reload-test`, because without the dependency on the latest resources lib --- there is no annotation in the class path and it fails.

AbdullinAM: And we don't want to add a dependency on the latest resource lib, because then it will pollute the whole classpath for tests

Issue #470, opened 2026-01-28
Gradle plugin eagerly realizes tasks during configuration phase


dovchinnikov:
**To Reproduce**

- Add an empty project with hot reload plugin:
```
plugins {
    kotlin("jvm") version "2.1.0"
    kotlin("plugin.compose") version "2.1.0"
    id("org.jetbrains.compose") version "1.7.3"
    id("org.jetbrains.compose.hot-reload") version "1.1.0-alpha04"
    application
    `java-test-fixtures`
}

tasks.configureEach {
    println("[REALIZED] $path")
}
```
- run `./gradlew` without arguments.
- observe output 
```
[REALIZED] :hotReloadDev
[REALIZED] :hotReloadMain
[REALIZED] :hotReloadTest
[REALIZED] :hotReloadTestFixtures
[REALIZED] :hotRun
[REALIZED] :hotSnapshotDev
[REALIZED] :hotSnapshotMain
[REALIZED] :hotSnapshotTest
[REALIZED] :hotSnapshotTestFixtures
[REALIZED] :reload
[REALIZED] :hotDev
[REALIZED] :hotRunArgfile
[REALIZED] :hotDevArgfile
[REALIZED] :hotRunAsync
[REALIZED] :hotDevAsync
```

**Expected behavior**

Hot reload tasks should be registered lazily and only realized when they are part of the task graph for execution.

**Environment**

Compose Hot Reload version: 1.1.0-alpha04

AbdullinAM:
cc @sellmair 

Did we have a reason for the eager configuration? I remember we talked about it, but I don't remember the reasoning

AbdullinAM:
Fixed in #475 by @dovchinnikov

PR #471, opened 2026-01-28
#470 fix eager task realization

dovchinnikov:
- Replace `name in tasks.names` checks with `try-catch` around `tasks.register()`
- Removed unnecessary duplicate checks.
- Replace `.all {}` with `.configureEach {}`.
- Replace `tasks.withType<T> {}` with `tasks.withType<T>().configureEach {}`.

### DIFF
--- a/hot-reload-gradle-core/src/main/kotlin/org/jetbrains/compose/reload/gradle/utils.kt
+++ b/hot-reload-gradle-core/src/main/kotlin/org/jetbrains/compose/reload/gradle/utils.kt
@@ -114,11 +114,11 @@ suspend fun <T> Project.forAllJvmCompilations(block: suspend (compilation: Kotli
     val futures = mutableListOf<Future<T>>()
 
     withKotlinPlugin {
-        kotlinJvmOrNull?.target?.compilations?.all { compilation ->
+        kotlinJvmOrNull?.target?.compilations?.configureEach { compilation ->
             futures += future { block(compilation) }
         }
-        kotlinMultiplatformOrNull?.targets?.withType(KotlinJvmTarget::class.java)?.all { target ->
-            target.compilations.all { compilation ->
+        kotlinMultiplatformOrNull?.targets?.withType(KotlinJvmTarget::class.java)?.configureEach { target ->
+            target.compilations.configureEach { compilation ->
                 futures += future { block(compilation) }
             }
         }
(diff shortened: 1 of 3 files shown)


hot-reload-gradle-core/src/main/kotlin/org/jetbrains/compose/reload/gradle/utils.kt
```diff
@@ -114,11 +114,11 @@ suspend fun <T> Project.forAllJvmCompilations(block: suspend (compilation: Kotli
     val futures = mutableListOf<Future<T>>()
 
     withKotlinPlugin {
-        kotlinJvmOrNull?.target?.compilations?.all { compilation ->
+        kotlinJvmOrNull?.target?.compilations?.configureEach { compilation ->
```
sellmair, line 117: Compilations, Targets, and other KGP models are not lazy; the semantics here are to hook directly into the creation and configuration of further entities.

hot-reload-gradle-plugin/src/main/kotlin/org/jetbrains/compose/reload/gradle/tasksBase.kt
```diff
@@ -31,8 +31,8 @@ interface ComposeHotReloadRunTask : ComposeHotTask
 interface ComposeHotReloadOtherTask : ComposeHotTask
 
 internal fun Project.configureComposeHotReloadTasks() {
-    tasks.withType<ComposeHotTask> {
-        group = when (this) {
+    tasks.withType<ComposeHotTask>().configureEach {
+        group = when (it) {
```
sellmair, line 35: I would assume that this is the change you need, right?
If all other changes are reverted, but only this is kept, then the desired effect should show. 

Would you mind creating a test for the compliance, or shall we take care for that?

sellmair:
@AbdullinAM @pjBooms : I think the MR requires taking over by the team.

AbdullinAM:
@sellmair Yes, thank you! I will finalise it

AbdullinAM:
Moved the original commits to #475 and finalised them, as I could not add/change commits directly in this PR. #475 is merged, so I'm closing this one.

@dovchinnikov Thanks for the contribution!

dovchinnikov:
Thank you for taking over this PR!

Issue #469, opened 2026-01-26
Hovering the mouse cursor over the dev tools window crashes the application


Nohus:
**Describe the bug**

Then application crashes when the dev tools windowed is hovered over:

https://github.com/user-attachments/assets/9f3d410c-3fb4-4bc0-94db-7cdfd298e0e9

**To Reproduce**

Hover over the dev tools window

**Expected behavior**

Not crashing

**Environment**
 - OS: Kubuntu 24.04
 - Kotlin version: `2.3.0`
 - Compose Multiplatform version: `1.10.0`
 - Compose Hot Reload version: `1.1.0-alpha04`

**Stack Trace**
```
Application Orchestration closed
java.io.EOFException
	at java.base/java.io.DataInputStream.readFully(DataInputStream.java:215)
	at java.base/java.io.DataInputStream.readInt(DataInputStream.java:390)
	at org.jetbrains.compose.reload.orchestration.OrchestrationIOImpl$readPackage$2.invokeSuspend(OrchestrationIO.kt:106)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
	at org.jetbrains.compose.reload.core.WorkerThreadDispatcher$interceptContinuation$1$future$1.invoke(WorkerThreadDispatcher.kt:31)
	at org.jetbrains.compose.reload.core.WorkerThreadDispatcher$interceptContinuation$1$future$1.invoke(WorkerThreadDispatcher.kt:30)
	at org.jetbrains.compose.reload.core.WorkerThread.run$execute(WorkerThread.kt:25)
	at org.jetbrains.compose.reload.core.WorkerThread.run(WorkerThread.kt:42)
	Suppressed: org.jetbrains.compose.reload.core.AsyncTracesThrowable: Async Traces
	Caused by: org.jetbrains.compose.reload.core.AsyncTracesThrowable$FrameThrowable: subtask(Reader)
		at org.jetbrains.compose.reload.core.AsyncTraces$Frame.<init>(AsyncTraces.kt:29)
		at org.jetbrains.compose.reload.core.TaskImpl.subtask(Task.kt:305)
		at org.jetbrains.compose.reload.core.Task$DefaultImpls.subtask$default(Task.kt:29)
		at org.jetbrains.compose.reload.orchestration.OrchestrationClientKt$OrchestrationClient$task$1.invokeSuspend(OrchestrationClient.kt:188)
[...]

PR #472, opened 2026-02-02
Fix: use the same transparency in the tooltip as in the sidecar

AbdullinAM:
Fix for #469

### DIFF
--- a/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtTooltip.kt
+++ b/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtTooltip.kt
@@ -11,6 +11,7 @@ import androidx.compose.foundation.hoverable
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.interaction.collectIsHoveredAsState
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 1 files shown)


hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtTooltip.kt
```diff
@@ -46,6 +48,11 @@ private val tooltipHideDelay = 100.milliseconds
 
 private val defaultTooltipOffset = DpSize(1.dp, 1.dp)
 
+private val tooltipCornerShape = when (devToolsUseTransparency) {
+    true -> DtShapes.TooltipCornerShape
```
pjBooms: I would not use `when` for boolean values. Why `if` does not work here?

AbdullinAM: I stylistically always prefer `when` over `if`, because I think that it is better for readability. I updated the condition to ensure that it compiles the same as regular if

pjBooms: > because I think that it is better for readability

Do you think Kotlin should deprecate `if` statement? For me `if` is better for readability and more compact. I used to use `if` for years to read them well.

AbdullinAM: I don't think deprecating `if` statement is a good solution, because there are many people that like and use it. I'm just not one of them. 

Comparing
```kotlin
val a = when {
    b -> ...
    else -> ...
}
```

To
```kotlin
// (1)
val a = if (b) ... else ...
// (2)
val a = if (b) ...
else ...
```

I like `when` better, because:

* each branch is clearly separated by a new line(s)
* there are clean anchor points to quickly locate both the condition and the body

I only try to use (1) if the statement is short (fits in 80 symbol column)

PR #473, opened 2026-02-02
Custom JBR provisioning for hot reload tasks

AbdullinAM:
New experimental property to enable automatic JBR provisioning for hot reload Gradle tasks

### DIFF
--- a/hot-reload-gradle-core/src/main/kotlin/org/jetbrains/compose/reload/gradle/JbrProvisioner.kt
+++ b/hot-reload-gradle-core/src/main/kotlin/org/jetbrains/compose/reload/gradle/JbrProvisioner.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2024-2026 JetBrains s.r.o. and Compose Hot Reload contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ */
+
+@file:Suppress("UnstableApiUsage")
+
+package org.jetbrains.compose.reload.gradle
+
+import org.jetbrains.compose.reload.core.Arch
+import org.jetbrains.compose.reload.core.JavaHome
+import org.jetbrains.compose.reload.core.JbrMetadata
+import org.jetbrains.compose.reload.core.LockFile
+import org.jetbrains.compose.reload.core.Os
+import java.io.FileOutputStream
+import java.nio.file.Files
+import java.nio.file.Path
+import kotlin.io.path.*
+import org.gradle.api.file.ArchiveOperations
+import org.gradle.jvm.toolchain.JavaLanguageVersion
+import org.jetbrains.compose.reload.InternalHotReloadApi
+import java.net.URL
+
+@InternalHotReloadApi
+class JbrProvisioner(
+    private val gradleUserHome: Path,
+    private val archiveOperations: ArchiveOperations
+) {
+    fun provision(java: JavaLanguageVersion): JavaHome? {
+        val jbrVersion = JbrMetadata.getVersionCompatibleWith(java.asInt())
+        val os = Os.currentOrNull() ?: return null
+        val arch = Arch.currentOrNull() ?: return null
+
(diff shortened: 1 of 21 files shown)


hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/JbrMetadata.kt
```diff
@@ -0,0 +1,52 @@
...
+ * Shared between the Compose Hot Reload Gradle plugin and Compose Hot Reload tests plugin to maintain consistency.
+ */
+@InternalHotReloadApi
+public object JbrMetadata {
+    public const val BASE_URL: String = "https://cache-redirector.jetbrains.com/intellij-jbr"
+    public const val VENDOR_NAME: String = "JetBrains"
+
+    @InternalHotReloadApi
+    public enum class Version(public val version: String, public val build: String) {
+        JBR_25("25", "b176.4"),
```
pjBooms: Does it mean that we hardwire builds in the code? 
When it was in tests only it looked ok, but now it becomes more public.
Will we have any policy to update this metadata?

hot-reload-gradle-core/src/main/kotlin/org/jetbrains/compose/reload/gradle/jetbrainsRuntime.kt
```diff
@@ -29,20 +31,24 @@ import kotlin.io.path.absolutePathString
...
  *
  * 2. Gradle provisioned
  * This provisioning can even download the suitable JBR (using the foojay resolver).
  * It uses the [JavaToolchainService] to request a JetBrains runtime using a suitable languageVersion.
  * This language version matches the project level toolchain's language version or falls back to the
- * [org.jetbrains.compose.reload.core.HotReloadProperty.JetBrainsRuntimeVersion]
+ * [HotReloadProperty.JetBrainsRuntimeVersion]
  *
- * 3. IntelliJ provided
+ * 3. Automatically provisioned
```
pjBooms, line 43: I would update README.md with this info as well.

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/JbrMetadata.kt
AbdullinAM: Yes, that is a good question. 

I had ideas to either get this info automatically, or load it from the separate file.
This first option will ideally not require any maintenance and updating, but will potentially be more flaky/prone to errors, because we will have to rely on external APIs to get the latest versions.

Loading from file seems good, and probably the best option. But it will still be relatively 'hardcoded', meaning we won't be able to update JBR metadata for already released versions. But I don't think it should be critical, this is still meant to be a safe fallback option.

> Will we have any policy to update this metadata?

Same as with other versions, someone will have to do it manually. Thats why I think loading metadata from file will be better, we can store it somewhere close to the other version files

hot-reload-gradle-core/src/main/kotlin/org/jetbrains/compose/reload/gradle/JbrProvisioner.kt
```diff
@@ -0,0 +1,91 @@
...
+) {
+    fun provision(java: JavaLanguageVersion): JavaHome? {
+        val jbrVersion = JbrMetadata.getVersionCompatibleWith(java.asInt())
+        val os = Os.currentOrNull() ?: return null
+        val arch = Arch.currentOrNull() ?: return null
+
+        val dirName = JbrMetadata.getJbrFileNameFor(jbrVersion, os, arch)
+        val cachedJbrDir = gradleUserHome
+            .resolve(CACHE_DIR)
+            .resolve(dirName)
```
sellmair, line 37: ~/.gradle/caches might not be the best place, as it requires the question if this can be relocated across machines. Maybe using ~/.gradle/chr/jbr or something like this would be more 'inline' with how others work with this directory

PR #479, opened 2026-02-09
[logging] Display more of recompiler error messages with 'Error' level

AbdullinAM:


AbdullinAM:
> It improves the logging of course, but makes me wonder if we can fix the root cause: understand what gradle report as an error in its output.

We can, in theory, switch the recompiler to `--console=rich` and try to parse/match the Gradle rich output. But I don't really like that idea

PR #480, opened 2026-02-09
CMP-9732: restart from tool window bug fix.

pjBooms:
Fixes: [CMP-9732](https://youtrack.jetbrains.com/issue/CMP-9732/Restart-button-in-Hot-Reload-Tool-Window-returns-the-app-to-the-state-before-the-changes)

After investigation, I noticed that `hot` directory is not exclusively owned by `hotSnapshot` task because `hotRun` task actually cleans it on the app start. 
So marking it `@OutputDirectory` seems to be not correct.

On the other hand, I have not noticed problems marking it as `@Internal` and it fixes the restart case.


sellmair:
> On the other hand, I have not noticed problems marking it as @Internal and it fixes the restart case.

Fair, however, I advise that extensive testing is performed here, trying to break the up-to-dateness and consistency of reloads within multiple builds.

PR #481, opened 2026-02-09
readme: min requirements, explicit and auto modes

ElviraMustafina:


README.md
```diff
@@ -83,34 +87,57 @@ guide to set up your environment and create a project. Be sure to select the des
    ```
     Alternatively, you can use automatic JetBrains Runtime provisioning for the hot reload tasks by enabling
     the `compose.reload.jbr.autoProvisioningEnabled` property.
-> [!IMPORTANT]  
-> Automatic JetBrains Runtime provisioning is an experimental feature. Please report any issues you encounter.
+
+    > [!IMPORTANT]
```
AbdullinAM: The rendering of this breaks if you add spaces in front

Image

README.md
```diff
@@ -83,34 +87,57 @@ guide to set up your environment and create a project. Be sure to select the des
...
+
+### Compose Hot Reload modes
+
+In **Explicit** mode, you need to manually trigger the reload to apply new changes.
+To trigger a reload, you can use any of the following options:
 
-### From the IDE
+* Click the **Reload** button in the IDE.
+* Run the `./gradlew reload` task in the CLI.
+* Configure reloads to trigger on certain actions, such as saving, pressing a hotkey, or detecting external changes.
```
AbdullinAM: I think this description makes the explicit mode sound more complicated than it is for the user. It sounds like the user needs to configure something before using the explicit mode in IDE without giving enough context.

Any IDE integration (including the 'settings | tools | CHR' page and 'reload' button) for CHR require having KMP installed. If KMP is installed, the 'explicit' mode will behave similar to the 'auto', because KMP is configured by default to trigger the reload on 'Save All'.
Image

So I think it would be nice to provide description on how explicit mode works for all three cases:
* Running in IDE with KMP
* Running in IDE without KMP
* Running from CLI

README.md
```diff
@@ -83,34 +87,57 @@ guide to set up your environment and create a project. Be sure to select the des
...
+You can run your application with Compose Hot Reload using your IDE or the CLI via Gradle tasks.
+Compose Hot Reload supports two modes: **Explicit** mode and **Auto** mode.
+
+### Compose Hot Reload modes
+
+In **Explicit** mode, you need to manually trigger the reload to apply new changes.
+To trigger a reload, you can use any of the following options:
 
-### From the IDE
+* Click the **Reload** button in the IDE.
```
AbdullinAM: You can also use 'reload' button in the dev tools window. This is how it looks in 1.1, the UI is a bit different in 1.0.0

Image

README.md
```diff
@@ -83,34 +87,57 @@ guide to set up your environment and create a project. Be sure to select the des
...
+* Run the `./gradlew reload` task in the CLI.
+* Configure reloads to trigger on certain actions, such as saving, pressing a hotkey, or detecting external changes.
+  You can choose the preferred triggers on the **Settings | Tools | Compose Hot Reload** page in your IDE.
 
-In IntelliJ IDEA or Android Studio, in the gutter, click the **Run**
-icon <img alt="Run main function" src="./readme-assets/run.png" width="12"> of your main function:
+In **Auto** mode, Compose Hot Reload uses Gradle’s file-watching and continuous build system
+to automatically reload when file changes are detected.
+It is enabled by default when running from the CLI. To enable this mode in the IDE,
```
AbdullinAM: This is not true, `--auto` mode is not enabled by default anywhere. We always launch in explicit mode by default

README.md
```diff
@@ -83,34 +87,57 @@ guide to set up your environment and create a project. Be sure to select the des
...
+
+In IntelliJ IDEA or Android Studio, you can run Compose Hot Reload directly from the IDE's gutter.
+
+If you have the [Kotlin Multiplatform IDE plugin](https://plugins.jetbrains.com/plugin/14936-kotlin-multiplatform)
+installed, click the **Run** icon <img alt="Run main function" src="./readme-assets/run.png" width="12"> 
+in the gutter of your main function and select **Run 'shared [jvm]' with Compose Hot Reload**.
+Otherwise, create Gradle run configurations with hot reload tasks manually (see [Run tasks](#run-tasks)).
+
+By default, Compose Hot Reload in the IDE runs in Explicit mode. 
+To enable Auto mode, specify the `--autoReload` (or `--auto`) argument.
```
AbdullinAM: Our goal with the IDE experience for 'explicit' mode was to make as easy as '--auto' mode. So that the users don't even notice that they actually are running in explicit mode. Maybe it's not worth explicitly mentioning auto mode in IDE, it will just confuse users more.

README.md
```diff
@@ -83,34 +87,57 @@ guide to set up your environment and create a project. Be sure to select the des
...
 
 ### From the CLI
 
-#### Run tasks
+By default, Compose Hot Reload runs from the CLI in Auto mode with `hotRun*` tasks.
```
AbdullinAM: Same as above

README.md
```diff
@@ -83,34 +87,57 @@ guide to set up your environment and create a project. Be sure to select the des
...
+To use Explicit mode, you can run reloading tasks manually.
+
+#### Run tasks (Auto mode)
 
-The Compose Hot Reload plugin automatically creates the following tasks to launch the application in 'hot reload mode':
+The Compose Hot Reload plugin automatically creates the following tasks to launch the application in Auto mode:
 
 - `:hotRunJvm`: For multiplatform projects. The async alternative is `:hotRunJvmAsync`.
```
AbdullinAM, line 139: These tasks are available for both auto and explicit mode. The original description suits well here I think: these are the tasks that CHR generates for your project so that you can run it with hot reload enabled.

AbdullinAM (changes requested):
Thanks for updating the docs! I like the more detailed descriptions, but there are a few technical details that need to be fixed

README.md
ElviraMustafina: Oops!

README.md
ElviraMustafina: All set! Should be way better now

README.md
```diff
@@ -24,13 +24,17 @@ The JetBrains Runtime intelligently reloads your code whenever it changes.
...
 - Kotlin 2.1.20 or higher.
 - Compose compiler 2.1.20 or higher.
-- Compose Multiplatform 1.8.2
-- [JetBrains Runtime](https://github.com/JetBrains/JetBrainsRuntime).
+- Compose Multiplatform 1.8.2 or higher.
+- [JetBrains Runtime](https://github.com/JetBrains/JetBrainsRuntime):
   To be compatible with JetBrains Runtime, your project needs
   to [target](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation)
   Java 21 or earlier.
+- IntelliJ IDEA 2025.2.2 or higher, or Android Studio Otter 2025.2.1 or higher.
```
AbdullinAM, line 39: While I agree that IDEs provide the best experience with CHR, it is not a hard requirement. You can easily use CHR from CLI too, you just have to understand a bit more. Or, even if you are using IDE, KMP is not a hard requirement as well.

So I don't think we should mention IDE and KMP in prerequisites, at least not at the same level as "Kotlin 2.1.20+" and CMP 1.8.2+

README.md
```diff
@@ -92,25 +97,41 @@ guide to set up your environment and create a project. Be sure to select the des
...
+
+### In the IDE
+
+In IntelliJ IDEA or Android Studio, you can run Compose Hot Reload directly from the IDE gutter.
+
+1. Click the **Run** icon <img alt="Run main function" src="./readme-assets/run.png" width="12"> 
+in the gutter of your main function and select **Run 'shared [jvm]' with Compose Hot Reload**.
 
-### From the IDE
+2. After making changes to your code, click the **Reload UI** button:
```
AbdullinAM: I still think this is a bit confusing and potentially misleading:

If you have KMP installed, you don't need to manually click the 'reload ui' button. IDE will detect that you have saved changes to your file and trigger the reload itself. So from the user perspective, IDE experience with KMP is the same as auto mode. Just the monitoring done by IDE, not Gradle.

If you don't have KMP installed, then you won't have the run gutter configurations and you will need to create the Gradle run configs yourself.

So in this section point 1 and point 2 give instructions for different scenarios (IDE with KMP vs IDE without KMP).

README.md
ElviraMustafina: Moved IDE+KMP to the separate list with the recommended setup

README.md
ElviraMustafina: - Changed to 1. Run and 2. Save to reload
- Since the IDE+KMP setup is considered recommended, it is described as the default in the IDE section, with a mention of manually creating Gradle run configurations at the end.

README.md
```diff
@@ -92,25 +100,42 @@ guide to set up your environment and create a project. Be sure to select the des
...
+### In the IDE
 
-In IntelliJ IDEA or Android Studio, in the gutter, click the **Run**
-icon <img alt="Run main function" src="./readme-assets/run.png" width="12"> of your main function:
+In IntelliJ IDEA or Android Studio, you can run Compose Hot Reload directly from the IDE gutter.
 
-* If you have the [Kotlin Multiplatform IDE plugin](https://plugins.jetbrains.com/plugin/14936-kotlin-multiplatform)
-  installed, select **Run 'MainKt [hotRunJvm]' with Compose Hot Reload (Rc)**.
-* Otherwise, manually create Gradle run configurations with hot reload tasks (see [Run tasks](#run-tasks)).
+1. Click the **Run** icon <img alt="Run main function" src="./readme-assets/run.png" width="12">
```
AbdullinAM, line 116: Couple of small notes that come to mind. These are not hard requirements, just want to know your opinion on whether we should add them

* Should we still explicitly specify that this experience requires KMP?Or do you think that mentioning it in the note below is enough?
* Should we also mention `Settings -> Tools -> Compose Hot Reload' menu so that people know how to configure hotkeys for reload actions? Especially that we mention 'you can trigger the reload explicitly by pressing the assigned shortcut key' and don't explicitly say what is this shortcut key

README.md
```diff
@@ -92,25 +100,42 @@ guide to set up your environment and create a project. Be sure to select the des
...
+
+   Alternatively, you can trigger the reload explicitly by pressing the assigned shortcut key or
+   clicking the **Reload UI** button:
+
+    <img src="/readme-assets/compose-hot-reload-floating-toolbar.png" alt="Reload UI in the IDE" width="528">
+
+> [!IMPORTANT]  
+> If you don't have the [Kotlin Multiplatform IDE plugin](https://plugins.jetbrains.com/plugin/14936-kotlin-multiplatform)
+> installed, you can manually create Gradle run configurations with hot reload tasks (see [Run tasks](#run-tasks)).
+> In this case, you can trigger the reload by saving all files.
```
AbdullinAM: Correction: without KMP plugin the reload will not trigger on file save actions; you will need to manually press 'reload UI' button or execute a 'reload' gradle task. Or `--auto` mode

ElviraMustafina: Fixed!

README.md
ElviraMustafina: - Since we recommend the IDE+KMP setup in the prerequisites, I think it's okay to refer to it by default without extra details.
- Added info about the settings page.

AbdullinAM: Nice, thank you! I will squash the commit and merge

AbdullinAM (approved):

PR #512, opened 2026-04-08
Introduce MCP server POC

pjBooms:
## Summary

See: [CMP-10042](https://youtrack.jetbrains.com/issue/CMP-10042)

- Add `hot-reload-mcp` module — a [Model Context Protocol](https://modelcontextprotocol.io/) server that enables AI agents to interact with a running Compose application
- Expose `status` and `take_screenshot` tools over stdio transport
- The MCP server starts independently of the application, waits for it to launch via PID file polling, detects shutdown, and reconnects automatically
- Add `CLAUDE.md` with project overview and architecture documentation

## New module

| Module | Purpose |
|--------|---------|
| `hot-reload-mcp` | MCP server with `status` and `take_screenshot` tools, launched via `hotMcpServer` Gradle task |

## How it works

1. The Gradle plugin registers a `hotMcpServer` task that launches the MCP server process
2. The server polls for a PID file written by the running application to discover the orchestration port
3. Once connected, AI agents can call `status` to check availability and `take_screenshot` to capture the application window
4. If the application shuts down, the server detects the disconnect and waits for a new instance

## Testing

- [x] `McpServerTest` — 5 tests verifying MCP tool behavior via in-process piped client/server (status connected/disconnected, screenshot success/error, connection transitions)
- [x] `ConnectionLifecycleTest` — 4 tests verifying PID file polling and orchestration connect/disconnect/reconnect with real orchestration servers
- [x] `TakeScreenshotIntegrationTest` — screenshot capture in non-headless singleWindowApplication launch
- [x] `ComposeHotMcpServerTasksTest` — Gradle task registration


CLAUDE.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## Key Modules
+
+| Module | Purpose |
+|--------|---------|
+| `hot-reload-core` | Core hot reload functionality |
+| `hot-reload-gradle-plugin` | Gradle plugin (shaded for plugin portal) |
+| `hot-reload-gradle-core` | Core Gradle integration |
+| `hot-reload-gradle-idea` | IntelliJ IDEA / Android Studio integration |
```
AbdullinAM: This is the project/build model that is used by IDEs, not IDE integration itself

CLAUDE.md
```diff
@@ -0,0 +1,107 @@
...
+| `hot-reload-agent` | JVM agent that reloads classes on demand |
+| `hot-reload-orchestration` | Communication layer between Gradle, app, and IDE |
```
AbdullinAM: and devtools

CLAUDE.md
```diff
@@ -0,0 +1,107 @@
...
+| `hot-reload-runtime-jvm` | JVM runtime implementation |
```
AbdullinAM: This is a very unclear description of the module

CLAUDE.md
```diff
@@ -0,0 +1,107 @@
...
+| `hot-reload-runtime-api` | Public runtime API |
+| `hot-reload-annotations` | Kotlin annotations for hot reload API |
+| `hot-reload-analysis` | Bytecode analysis for reload compatibility |
```
AbdullinAM: Compatibility?

CLAUDE.md
```diff
@@ -0,0 +1,107 @@
...
+| `hot-reload-devtools` | Developer tools UI window |
+| `hot-reload-mcp` | MCP server for AI agent integration |
+| `hot-reload-test` | Test infrastructure and test plugins |
+| `tests/` | Functional integration tests |
+| `samples/` | Example projects (counter app) |
+| `buildSrc/` | Build conventions and custom Gradle tasks |
+
+## Architecture
+
+**Orchestration** — Communication between Gradle, the Application, and the IDE. The application starts an orchestration server; all connected clients exchange broadcast messages (e.g., `ReloadClassesRequest`).
```
AbdullinAM: The devtools is starting the orchestration server, not the app itself

README.md
```diff
@@ -229,6 +229,58 @@ dependencyResolutionManagement {
 }
 ```
 
+## MCP server for AI agents
+
+*Note: the MCP server is in experimental status
```
AbdullinAM: If we merge this, it will be available in the dev builds and later in the alpha releases too. So I don't see a lot of sense in the second part of this sentence

AbdullinAM: If we merge this, it will be available in the dev builds and later in the alpha releases too. So I don't see a lot of sense in the second part of this sentence

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlin.time.Duration.Companion.seconds
+
+private val logger = createLogger()
+
+fun main(args: Array<String>) {
+    val pidFilePath = args.firstOrNull()
+        ?: System.getProperty(HotReloadProperty.PidFile.key)
+        ?: run {
+            logger.error("PID file path not provided. Pass it as argument or via -D${HotReloadProperty.PidFile.key}=PATH")
+            System.exit(1)
```
AbdullinAM:
```kotlin
System.exit(1)
return
```

Can be replaced with single
```kotlin
exitProcess(1)
```

Smart casts will understand that the `pidFilePath` is never null

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,100 @@
...
+    val transport = StdioServerTransport(
+        inputStream = System.`in`.asSource().buffered(),
+        outputStream = System.out.asSink().buffered()
+    )
+    startMcpServer(orchestration, transport)
+}
+
+internal suspend fun startMcpServer(orchestration: StateFlow<OrchestrationHandle?>, transport: Transport) {
+    val server = Server(
+        serverInfo = Implementation(name = "compose-hot-reload", version = "1.0.0"),
```
AbdullinAM: I think the version here should be the same as the current CHR version. I would use `HOT_RELOAD_VERSION` constant

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+    val pidFile = Path.of(pidFilePath)
+    logger.info("MCP server starting. Watching PID file: $pidFile")
+
+    val orchestration = MutableStateFlow<OrchestrationHandle?>(null)
+
+    runBlocking {
+        launch(Dispatchers.IO) {
+            connectionLoop(pidFile, orchestration)
```
AbdullinAM: Intuitively it makes more sense to me to have one function that produces a flow and another one that consumes it. Right now you can only guess which of the functions is the producer based on the function names, but not on the code structure. WDYT?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
```diff
@@ -0,0 +1,100 @@
...
+ * Continuously watches the PID file for an orchestration port.
+ * When the port is found, connects to the orchestration server.
+ * When the connection drops, goes back to polling the PID file.
+ */
+internal suspend fun connectionLoop(pidFile: Path, orchestration: MutableStateFlow<OrchestrationHandle?>) {
+    while (true) {
+        val port = pollForOrchestrationPort(pidFile)
+        logger.info("Found orchestration port $port, connecting...")
+
+        val client = OrchestrationClient(OrchestrationClientRole.Tooling, port)
```
AbdullinAM: You can use `connectOrchestrationClient` here

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
```diff
@@ -0,0 +1,100 @@
...
+            System.exit(1)
+            return
+        }
+
+    val pidFile = Path.of(pidFilePath)
+    logger.info("MCP server starting. Watching PID file: $pidFile")
+
+    val orchestration = MutableStateFlow<OrchestrationHandle?>(null)
+
+    runBlocking {
```
AbdullinAM, line 50: I would probably use the `launchTask` APIs from the chr concurrency model, just to be consistent with the other code

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,100 @@
...
+            CallToolResult(
+                content = listOf(TextContent("""{"connected": $connected}"""))
+            )
+        }
+
+        addTool(
+            name = "take_screenshot",
+            description = "Take a screenshot of the running Compose application window. " +
+                "Use the 'status' tool first to check if the application is connected."
+        ) { _ ->
```
AbdullinAM, line 72: The potential problem with these implementations is that they may crash with an exception if the app is stopped in the middle of the tool's execution.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
```diff
@@ -0,0 +1,100 @@
...
+ * When the connection drops, goes back to polling the PID file.
+ */
+internal suspend fun connectionLoop(pidFile: Path, orchestration: MutableStateFlow<OrchestrationHandle?>) {
+    while (true) {
+        val port = pollForOrchestrationPort(pidFile)
+        logger.info("Found orchestration port $port, connecting...")
+
+        val client = OrchestrationClient(OrchestrationClientRole.Tooling, port)
+        val result = client.connect()
+        if (result.isFailure()) {
```
AbdullinAM, line 69: This can easily get into an infinite loop if the pid file is left undeleted/connection fails for unknown reasons. Should we handle it somehow? Set a break condition?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,100 @@
...
+                    content = listOf(TextContent("No application is currently connected. Use the 'status' tool to check availability.")),
+                    isError = true
+                )
+
+            // asChannel() registers the listener eagerly before we send the request,
+            // avoiding the race condition where the Screenshot reply arrives before
+            // the cold asFlow() starts collecting.
+            val screenshotChannel = handle.asChannel()
+
+            handle.send(TakeScreenshotRequest())
```
AbdullinAM: In the current request/result API design of orchestration messages the responses always include an ID of the request so that we can match the result to the request precisely. I would try to keep this design

hot-reload-test/screenshot/src/main/kotlin/org/jetbrains/compose/reload/test/screenshot/checkScreenshot.kt
```diff
@@ -0,0 +1,90 @@
...
+import kotlin.io.path.nameWithoutExtension
+import kotlin.io.path.writeBytes
+import kotlin.test.fail
+
+/**
+ * Compares [actualData] against the expected screenshot at [expectFile].
+ * If [expectFile] does not exist, it is created from [actualData] and the test fails with a "Generated" message.
+ * If it exists, the images are compared with the given tolerance; on mismatch an "-actual" file is written and the test fails.
+ */
+public fun checkScreenshot(
```
AbdullinAM: Visibility and annotations on some of the checkScreenshot functions is lost, please check that they are migrated properly

hot-reload-runtime-jvm/src/test/kotlin/org/jetbrains/compose/reload/jvm/tests/ScreenshotHandlerTest.kt
```diff
@@ -0,0 +1,76 @@
+/*
```
AbdullinAM: Tbh I would suggest creating proper integration tests with the new 'TakeScreenshot' API.

* I think that extracting the screenshot module just for a single unit test is a little excessive
* Maybe we can use Compose API to create and manage screenshots? Then we will not depend on awt API, won't need to create windows in this test, and probably we will be able to migrate all tests to use the new screenshot API instead of purely test one? It seems like an extra work to maintain two separate screenshot mechanisms

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,100 @@
...
+
+internal suspend fun startMcpServer(orchestration: StateFlow<OrchestrationHandle?>) {
+    val transport = StdioServerTransport(
+        inputStream = System.`in`.asSource().buffered(),
+        outputStream = System.out.asSink().buffered()
+    )
+    startMcpServer(orchestration, transport)
+}
+
+internal suspend fun startMcpServer(orchestration: StateFlow<OrchestrationHandle?>, transport: Transport) {
```
AbdullinAM, line 49: I would suggest trying to refactor this function a little bit, because right now its a bit too much: it handles how the mcp server is created and registered and at the same time it also contains logic for each tool call handling. If we add more tools --- this will be very hard to understand

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,100 @@
...
+
+            // asChannel() registers the listener eagerly before we send the request,
+            // avoiding the race condition where the Screenshot reply arrives before
+            // the cold asFlow() starts collecting.
+            val screenshotChannel = handle.asChannel()
+
+            handle.send(TakeScreenshotRequest())
+
+            val screenshot = withTimeoutOrNull(10.seconds) {
+                screenshotChannel.receiveAsFlow().filterIsInstance<Screenshot>().first()
```
AbdullinAM: You can use `consumeAsFlow` here and then you won't have to manually close the channel

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
```diff
@@ -0,0 +1,100 @@
...
+            delay(2.seconds)
+            continue
+        }
+
+        logger.info("Connected to orchestration on port $port")
+        orchestration.value = client
+
+        // Block until the connection closes (app exit)
+        withContext(Dispatchers.IO) {
+            client.getBlocking()
```
AbdullinAM: Do we really need to block here? Why can't we just await until the client is closed?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
```diff
@@ -0,0 +1,100 @@
...
+        startMcpServer(orchestration)
+    }
+}
+
+/**
+ * Continuously watches the PID file for an orchestration port.
+ * When the port is found, connects to the orchestration server.
+ * When the connection drops, goes back to polling the PID file.
+ */
+internal suspend fun connectionLoop(pidFile: Path, orchestration: MutableStateFlow<OrchestrationHandle?>) {
```
AbdullinAM: Maybe we can use Java's file watch APIs to detect the file changes? Manual polling seems very ineffective and fragile

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.Screenshot
+import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.TakeScreenshotRequest
+import org.jetbrains.compose.reload.orchestration.asChannel
+import java.util.Base64
+import kotlin.time.Duration.Companion.seconds
+
+internal suspend fun startMcpServer(orchestration: StateFlow<OrchestrationHandle?>) {
+    val transport = StdioServerTransport(
+        inputStream = System.`in`.asSource().buffered(),
+        outputStream = System.out.asSink().buffered()
```
AbdullinAM, line 43: If server uses std streams as transport, will it potentially cause problems if we start writing logs to the stdout as well?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.jetbrains.compose.reload.orchestration.OrchestrationHandle
+import java.nio.file.Path
+import kotlin.io.path.isRegularFile
+import kotlin.time.Duration.Companion.seconds
+
+private val logger = createLogger()
+
+fun main(args: Array<String>) {
+    val pidFilePath = args.firstOrNull()
+        ?: System.getProperty(HotReloadProperty.PidFile.key)
```
AbdullinAM: Why don't we use standart `org.jetbrains.compose.reload.core.HotReloadEnvironment.pidFile` here? It will simplify the code and it will be more aligned with how `HotReloadProperties` are passed and acessed in other modules

hot-reload-mcp/src/test/kotlin/org/jetbrains/compose/reload/mcp/McpServerTest.kt
```diff
@@ -0,0 +1,150 @@
...
+            launch {
+                server.asFlow().filterIsInstance<TakeScreenshotRequest>().first()
+                server.send(Screenshot("png", ByteArray(4) { 0xFF.toByte() }))
+            }
+
+            val orchestration = MutableStateFlow<OrchestrationHandle?>(toolingClient)
+            val client = createMcpClient(orchestration)
+
+            val result = client.callTool("take_screenshot", emptyMap())
+            assertFalse(result.isError == true)
```
AbdullinAM: Small nitpick: this looks very weird, I would use `assertEquals` instead

AbdullinAM (changes requested):
Overall the code works, but I left several comments

I still have some concerns on whether the MCP server should be a part of the CHR, I will discuss them with Emil and Viktor too

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,100 @@
...
+    val server = Server(
+        serverInfo = Implementation(name = "compose-hot-reload", version = "1.0.0"),
+        options = ServerOptions(
+            capabilities = ServerCapabilities(
+                tools = ServerCapabilities.Tools(listChanged = false)
+            )
+        )
+    ) {
+        addTool(
+            name = "status",
```
AbdullinAM, line 59: Afterthought: Maybe this tool should return more information. E.g. reload status, failures, etc.

CLAUDE.md
pjBooms: fixed

CLAUDE.md
pjBooms: done

CLAUDE.md
pjBooms: Rephrased

CLAUDE.md
pjBooms: rephrased

CLAUDE.md
pjBooms: fixed

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
pjBooms: fixed

README.md
pjBooms: Ok, removed

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
pjBooms: done

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: Sure, fixed

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
pjBooms: done

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
pjBooms: done

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: added try/catch

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: refactored a little bit, please take a look

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: done

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/Main.kt
pjBooms: fixed

hot-reload-mcp/src/test/kotlin/org/jetbrains/compose/reload/mcp/McpServerTest.kt
pjBooms: used `assertNotEquals` (as quick fix suggested)

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: I guess we can do that in a next PR.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
pjBooms: So what bad can happen in this case? Instead of polling for pidfile in a loop, we will be connecting in a loop.  Once the application starts again, we read new port from the file that will break the loop. Anyway the server should be alive until an agent need it (an agent won't restart the server, if it stops).

hot-reload-runtime-jvm/src/test/kotlin/org/jetbrains/compose/reload/jvm/tests/ScreenshotHandlerTest.kt
pjBooms: I failed to find a proper Compose API for creating/managing screenshots except testing one. More precisely, there [is a way](https://github.com/android/snippets/blob/main/compose/snippets/src/main/java/com/example/compose/snippets/graphics/AdvancedGraphicsSnippets.kt#L92) to capture a Composable via GraphicsLayer but it is hardly fit to our case. `java.awt.Robot` is much simpler to use. Moreover `java.awt.Robot` also takes window title that may be useful for agent too. 

>and probably we will be able to migrate all tests to use the new screenshot API instead of purely test one

I doubt we can use the same mechanism both for tests and real screenshots taking. 

Regarding test, I wanted to have a test to see what actually we send to an agent. This way _expect_ screenshot appeared, that soon failed on other platforms. So the idea of re-using checkScreenshot API from `gradle-testFixtures` appeared and so on.     

>Tbh I would suggest creating proper integration tests with the new 'TakeScreenshot' API.

How the test could look like? I believe it should anyway launch a real application send a screenshot take request and after that we should compare again obtained screenshot with some expected one. Yes, we don't test the whole chain now, but test all separate parts. 

Perhaps, it is worth to discuss it in person, it is the most slippery part.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
pjBooms: Ok, re-implemented via WatchService, please take a look

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
pjBooms: refactored connectionLoop -- please take a look

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: Ok, I introduced new pair of messages to follow this. Please take a look.

sellmair:
Without reading the code:
If your MCP server always starts, then instead of checking the PID file, you can instruct the orchestration server to connect to your client. This machine would work similarly to how the IDE is instructing the application to connect to it once started.

hot-reload-test/screenshot/src/main/kotlin/org/jetbrains/compose/reload/test/screenshot/checkScreenshot.kt
pjBooms: I rolled back `screenshot` module

hot-reload-runtime-jvm/src/test/kotlin/org/jetbrains/compose/reload/jvm/tests/ScreenshotHandlerTest.kt
pjBooms: I added an integration test for the screenshots. The old unit-test for it was removed as superflous. Everything became simpler, imho.  Please take a look.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: I tested a bit and did not find problems: enabling logging writes logs messages while the server still responds ok.

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
```diff
@@ -301,6 +303,51 @@ internal constructor() : OrchestrationPackage(), Serializable {
...
+        }
+    }
+
+    /**
+     * Response to a [ScreenshotRequest], carrying the screenshot image data.
+     * @param screenshotRequestId the [OrchestrationMessageId] of the originating [ScreenshotRequest]
+     * @param format the image format (for example, 'png')
+     * @param data the raw image data
+     */
+    public class ScreenshotResult(
```
AbdullinAM, line 326: Do we have a scenario where the screenshot has failed? Maybe this should include some error info in that case?

AbdullinAM (commented):
I think we can publish a dev build in this state and then polish out the code after the feedback

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
pjBooms: Valid. Added an errorMessage to ScreenshotResult

pjBooms:
> Without reading the code: If your MCP server always starts, then instead of checking the PID file, you can instruct the orchestration server to connect to your client. This machine would work similarly to how the IDE is instructing the application to connect to it once started.

Unlike IDE, MCP server does not launch the app, so it cannot pass `-Dcompose.reload.orchestration.listener.port.` to it.

sellmair:
Also, please consult with @fkistner about this implementation.

CLAUDE.md
```diff
@@ -0,0 +1,107 @@
...
+       repositories { maven("file://path/to/compose-hot-reload/build/repo") }
+   }
+   ```
+5. Update the plugin version to match: `id("org.jetbrains.compose.hot-reload") version "1.0.0-DEBUG"`
+
+## Versioning
+
+- Current version: `gradle.properties` → `version`
+- Bootstrap version: `gradle.properties` → `bootstrap.version`
+- Main branch for PRs: `master`
```
AbdullinAM: `staging`

README.md
```diff
@@ -229,6 +229,57 @@ dependencyResolutionManagement {
 }
 ```
 
+## MCP server for AI agents
```
AbdullinAM, line 232: We will also want a review from @ElviraMustafina for the docs changes

AbdullinAM (approved):
Looks good, we can prepare a dev build with this version

We will need to clean up the commit history before merging

README.md
```diff
@@ -229,6 +229,57 @@ dependencyResolutionManagement {
...
+
+*Note: the MCP server is in experimental status.*
+
+Compose Hot Reload includes a [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) server that allows
+AI agents to interact with a running Compose application. The MCP server connects to the application's orchestration
+layer and exposes the following tools:
+
+| Tool | Description |
+|------|-------------|
+| `status` | Check whether a Compose application is currently connected |
```
ElviraMustafina: Minor suggestion: let's use "Checks" and "Captures" in the tool descriptions

ElviraMustafina (approved):

hot-reload-runtime-jvm/src/test/kotlin/org/jetbrains/compose/reload/jvm/tests/ScreenshotHandlerTest.kt
fkistner: > there [is a way](https://github.com/android/snippets/blob/main/compose/snippets/src/main/java/com/example/compose/snippets/graphics/AdvancedGraphicsSnippets.kt#L92) to capture a Composable via GraphicsLayer but it is hardly fit to our case

Why not? Would love to hear your thoughts on why it is not what we want.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
```diff
@@ -0,0 +1,118 @@
...
+private val logger = createLogger()
+
+@OptIn(InternalHotReloadApi::class)
+fun main(args: Array<String>) {
+    val pidFile = HotReloadEnvironment.pidFile
+        ?: args.firstOrNull()?.let { Path.of(it) }
+        ?: run {
+            logger.error("PID file path not provided. Set it via -Dcompose.reload.pidFile=PATH or pass as argument")
+            exitProcess(1)
+        }
```
fkistner, line 46: Instead of tying the server specifically to one pid, it would be great to support multiple Compose Hot Reload instances concurrently (different main classes, different Gradle modules, even multiple windows of the same project). Right now this resolution path allows exactly one app per MCP server, and two `hotRunJvm` invocations of the same compilation will silently clobber each other's pidFile.

What do you think about adopting a model where there is a `sessions` tool to list running Compose Hot Reload apps the agent can attach to, similar to [`playwright-mcp`](https://github.com/microsoft/playwright-mcp)'s `browser_tab_list` / `browser_tab_select` or [`chrome-devtools-mcp`](https://github.com/ChromeDevTools/chrome-devtools-mcp)'s `list_pages`: the MCP server watches the project's run directory (`<projectDir>/build/run/*/*.pid`, or `**/build/run/*/*.pid` for a multi-module build) and surfaces each entry as `{ id, projectPath, mainClass, pid, windowTitle?, connected }`. `status` and `take_screenshot` take an optional `session` argument and default to the only live one when there's exactly one.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    logger.info("MCP server starting. Watching PID file: $pidFile")
```
fkistner, line 48:
```
kotlin-logging: initializing... active logger factory: Slf4jLoggerFactory
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
```

Should we register a stderr logger so log messages can actually be seen?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/screenshotHandler.kt
```diff
@@ -0,0 +1,63 @@
...
+import java.awt.Rectangle
+import java.awt.Robot
+import java.awt.Window
+import java.awt.image.BufferedImage
+import java.io.ByteArrayOutputStream
+import javax.imageio.ImageIO
+
+private val logger = createLogger()
+
+internal fun handleScreenshotRequest(request: ScreenshotRequest, window: Window): ScreenshotResult {
```
fkistner, line 26: Does all of this need to run on EDT?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/screenshotHandler.kt
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * It captures the window title in addition to Compose content,
+ * so AI agent can analyze and understand the full context of the window.
+ */
+internal fun captureWindow(window: Window): Try<BufferedImage> {
+    return Try {
+        val robot = Robot()
+        val location = window.locationOnScreen
+        val rect = Rectangle(location.x, location.y, window.width, window.height)
+        robot.createScreenCapture(rect)
```
fkistner, line 61: What happens without macOS Screen Recording permission? Does it throw, or does it return a black image? Should that be handled?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -0,0 +1,124 @@
...
+            description = "Check whether a Compose application is currently connected. " +
+                "Returns {\"connected\": true} if the application is running and connected, " +
+                "or {\"connected\": false} if it is not. " +
+                "Call this before take_screenshot to know if the application is available."
```
fkistner, line 63: Maybe a `reason` field for communicating problems, would be nice?

fkistner:
Could it make sense to offer both MCP and command-line interfaces for the same tool surface? Recently, skills + cli has taken off (partly because it sidesteps MCP host registration entirely). The agent reads a `SKILL.md` where you can even include instructions about how to setup hot reload in the first place. Agents then just run the cli, get the output on stdout.

No `.mcp.json`, no JSON-RPC framing, no daemon, no version drift between an MCP host's bundled SDK and ours. For agents and runners on hosts without MCP support (Codex, raw Bash, GitHub Actions) it's the only path.

Mirroring the current tools (and the `sessions` proposal above), this could look like:

```bash
# List running Compose Hot Reload apps in this project
./gradlew :app:hotSessions
# → [{ "id": "jvmMain", "projectPath": "…", "mainClass": "…", "pid": 12345, "connected": true }, …]

# Connection status (defaults to the only live session)
./gradlew :app:hotStatus
./gradlew :app:hotStatus --session jvmMain

# Capture a screenshot
./gradlew :app:hotScreenshot --output screen.png
./gradlew :app:hotScreenshot --session jvmMain --output screen.png
```

Or this could be a standalone cli entirely (e.g., `kmp hotreload screenshot`, or similar).

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
pjBooms: Valid. Thanx! I have added a stderr logger in the last commit

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
pjBooms: I see that CHR now has a general problem on two instances simultaneously. I have created [an issue](https://youtrack.jetbrains.com/issue/CMP-10181/CHR-support-multiple-application-instances-running-with-hot-reload) on this. 

I don't think it is worth to address the issue in this PR but in general it is good idea to to support multiple Compose Hot Reload instances concurrently.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/screenshotHandler.kt
pjBooms: I am not sure what you mean, please refine.

`Robot.createScreenCapture()` is thread-safe and doesn't require EDT, but the current call site already runs on EDT via `LaunchedEffect.` Moving the capture off EDT is possible but unnecessary for our use case.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/screenshotHandler.kt
pjBooms: Hum. It works on my macOS even without Screen & System Audio Recording permission for IntelliJ IDEA.
Though it seems (from internet) it can return black image (I just can't reproduce it). 
I am not sure if it can be properly handled.
Taking into account that is is a first version, I think we can keep the current implementation and wait for the users feedback to return to the subject later.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: Status tool can be [improved](https://github.com/JetBrains/compose-hot-reload/pull/512/changes#r3085541814), indeed. I am going to handle it in next PRs

pjBooms:
> Could it make sense to offer both MCP and command-line interfaces for the same tool surface? Recently, skills + cli has taken off (partly because it sidesteps MCP host registration entirely). The agent reads a `SKILL.md` where you can even include instructions about how to setup hot reload in the first place. Agents then just run the cli, get the output on stdout.

That is a great idea but certainly not for this PR. I have created [an issue](https://youtrack.jetbrains.com/issue/CMP-10182) to keep it in mind.

hot-reload-runtime-jvm/src/test/kotlin/org/jetbrains/compose/reload/jvm/tests/ScreenshotHandlerTest.kt
pjBooms: I was not able to make use of it fast. I will experiment with it later.

PR #516, opened 2026-04-28
Update versions

AbdullinAM:
* Update JBR auto-provisioning versions
* Use JBR 25 as a default fallback, introduce `jbr.min.version` property
* Switch to Kotlin 2.3.21 for the CHR 1.2 release
* Fix compatibility issues with Kotlin 2.3.21 in tests
* Update test data
* Update test dimensions


vanniktech:
Reading `Fix compatibility issues with Kotlin 2.3.21` is 1.1.0 compatible with Kotlin 2.3.21 or should I contine to use 2.3.20?

AbdullinAM:
@vanniktech Yes, 1.1.0 is compatible with Kotlin 2.3.21. From the users POV, CHR is still compatible with all versions of Kotlin starting from 2.1.20.

The compatibility issues I mention here are related to the Compose Hot Reload's test infrastructure after updating its Kotlin version to 2.3.21. I updated the original comment to hopefully make it a bit clearer

properties.yaml
```diff
@@ -264,12 +264,21 @@ JetBrainsRuntimeBinary:
...
     Specifies the default 'JetBrains Runtime' version that shall be used (e.g. '21' or '25')
 
+JetBrainsRuntimeMinimalVersion:
+  key: compose.reload.jbr.min.version
+  type: int
+  default: "21"
+  target: [ build ]
+  visibility: internal
+  documentation: |
+    Specifies the minimal 'JetBrains Runtime' version that is supported
```
pjBooms: It is quite strange that a user can set minimal _supported_ version. Moreover it is unclear what it actually means. Looking at semantic of the property, I would say that it is also kind a default version that should be used when project language level is below of what we tested (support?).  Maybe it is better to name it as `compose.reload.jbr.version.lower.limit` and/or refine the documentation (I guess that documentation of  `compose.reload.jbr.version` should also be refined)

pjBooms (changes requested):

properties.yaml
AbdullinAM: Updated the docs

tests/src/reloadFunctionalTest/resources/screenshots/org/jetbrains/compose/reload/tests/ReloadEffectsTest/test - reload effects/2-reloading-500ms.png
AbdullinAM, line 1: The images themselves are still 100% the same, the difference is in size; updated images's sizes are 5-15 bytes lower. I think it comes from some differences in image handling/rendering after the updating versions

properties.yaml
```diff
@@ -264,11 +264,23 @@ JetBrainsRuntimeBinary:
...
+    user project does not explicitly declare toolchain requirements.
+
+JetBrainsRuntimeMinimalVersion:
+  key: compose.reload.jbr.min.version
+  type: int
+  default: "21"
+  target: [ build ]
+  visibility: internal
+  documentation: |
+    Specifies the minimal 'JetBrains Runtime' version that is supported. If the user
```
pjBooms: What about this variant:

```
Specifies the minimal 'JetBrains Runtime' version required by Compose Hot Reload.
    If the project's toolchain language version is set but is lower than this value,
    the plugin will override it and use this version instead.
```

pjBooms (approved):

properties.yaml
```diff
@@ -264,11 +264,22 @@ JetBrainsRuntimeBinary:
...
+    user project does not explicitly declare toolchain requirements.
+
+JetBrainsRuntimeMinimalVersion:
+  key: compose.reload.jbr.min.version
+  type: int
+  default: "21"
+  target: [ build ]
+  visibility: internal
+  documentation: |
+    Specifies the minimal 'JetBrains Runtime' version that is supported. If the project's toolchain version
```
pjBooms: What about this variant:

```
Specifies the minimal 'JetBrains Runtime' version required by Compose Hot Reload. 
If the project's toolchain language version is set but is lower than this value, 
the plugin will override it and use this version instead.
````

pjBooms (approved):

PR #522, opened 2026-05-14
[mcpserver] Add UI actions tools: click, type_text, scroll, etc.

pjBooms:
## Summary

Adds UI-action tools to the Compose Hot Reload MCP server so AI agents can drive a running Compose application, not just observe it. Built on top of the existing `get_semantic_tree` tool — agents read the tree to discover node ids and supported actions, then dispatch one of:

| Tool | Semantic action | Arguments |
|------|-----------------|-----------|
| `click` | `OnClick` | `nodeId` |
| `long_click` | `OnLongClick` | `nodeId` |
| `type_text` | `SetText` | `nodeId`, `text` |
| `scroll` | `ScrollBy` | `nodeId`, `deltaX`, `deltaY` |
| `scroll_to_index` | `ScrollToIndex` | `nodeId`, `index` |

Each tool returns only after the application has executed the action and acknowledged it (`UIActionResult`), so callers can chain a follow-up `get_semantic_tree` / `take_screenshot` to observe the effect.

## Implementation notes

- New orchestration messages `UIActionRequest` / `UIActionResult` and a `UIAction` sealed hierarchy (`Click`, `LongClick`, `SetText`, `ScrollBy`, `ScrollToIndex`), with encoders, golden serialization data, and protocol-version bump.
- `uiActionHandler.kt` in `hot-reload-runtime-jvm` resolves the target `SemanticsNode` by id and invokes the corresponding `SemanticsActions.*` lambda. Failure modes (missing node, action not supported by node, action lambda absent, lambda returns `false`) all map to descriptive `UIActionResult` errors.
- Dispatch is wired from both the windowed `DevelopmentEntryPoint` path and the headless `runHeadless.kt` path used by screenshot tests, so the tools are exercisable from functional tests without a real AWT window.

## Test plan

- [x] `UIActionIntegrationTest` — 6 cases: `click`, `long_click`, `type_text`, `scroll` (with pre/post assertions that `LazyColumn` actually scrolled), `scroll_to_index`, missing-node error
- [x] `McpServerTest` — per-tool tests asserting the correct `UIAction` subtype reaches the app
- [x] `:hot-reload-orchestration:check` — encoder roundtrip + serialVersionUID / protocol-classifier golden data
- [x] Manual smoke: configure agent via `.mcp.json` against KotlinConf app and navigate within it

## Documentation

Updated `README.md` MCP tool table with the five new tools.

### DIFF
--- a/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
+++ b/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
@@ -9,11 +9,14 @@ import io.modelcontextprotocol.kotlin.sdk.server.Server
 import io.modelcontextprotocol.kotlin.sdk.server.ServerOptions
 import io.modelcontextprotocol.kotlin.sdk.server.StdioServerTransport
 import io.modelcontextprotocol.kotlin.sdk.shared.Transport
+import io.modelcontextprotocol.kotlin.sdk.types.CallToolRequest
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolResult
 import io.modelcontextprotocol.kotlin.sdk.types.ImageContent
 import io.modelcontextprotocol.kotlin.sdk.types.Implementation
 import io.modelcontextprotocol.kotlin.sdk.types.ServerCapabilities
 import io.modelcontextprotocol.kotlin.sdk.types.TextContent
+import io.modelcontextprotocol.kotlin.sdk.types.ToolSchema
+import kotlinx.coroutines.channels.ReceiveChannel
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.flow.filterIsInstance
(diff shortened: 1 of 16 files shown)


hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -84,6 +99,65 @@ internal suspend fun startMcpServer(orchestration: StateFlow<OrchestrationHandle
...
+        }
+
+        addTool(
+            name = "type_text",
+            description = "Replace the content of a text field with the given text. " +
+                "The element must expose 'editableText' as reported by 'get_semantic_tree'.",
+            inputSchema = typeTextToolSchema()
+        ) { request ->
+            handleUIAction(orchestration, request) { args ->
+                val text = (args?.get("text") as? JsonPrimitive)?.content
```
AbdullinAM, line 128: One of my concerns is that we potentially keep growing the number of places where we manually create/parse jsons. Even though it is created to be manually constructible, when you have a 400-line file with lots of different json operations, it is hard to fully grasp what is happening. Especially now that we also hava a more complex message exchange protocol

I don't know if MCP sdk allows it, but I would love to simplify this file somehow and extract the whole protocol into some more understandable form. Maybe not in this PR, but in the future

AbdullinAM (approved):

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: Ok, I will try to simplify it. I don't tink that json operations are the problem but making the protocol more understandable is something good to do.

PR #528, opened 2026-05-27
[teamcity] Fix git push failing in CI due to host key verification failure

pjBooms:
.. by adding github.com to known hosts

### DIFF
--- a/.teamcity/builds/conventions/PushPrivilege.kt
+++ b/.teamcity/builds/conventions/PushPrivilege.kt
@@ -19,6 +19,8 @@ fun BuildSteps.setupGit() {
         scriptContent = """
                 git config user.email "compose-team@jetbrains.com"
                 git config user.name "JetBrains Compose Team"
+                mkdir -p ~/.ssh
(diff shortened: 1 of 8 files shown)


AbdullinAM (changes requested):
The fix itself LGTM. Three comments:
* I would not push the dev build version update versions manually. Just fix the CI and let it do the rest of the work
* The CI fixes should be done via `teamcity` branch, because that branch is used in TC for loading build configs. Pushing the fix to `staging` or `master` will not do anything. I would recommend pushing the fixing commit to `teamcity` branch directly and then opening a PR to `staging`, so that the build configs are consistent between branches (like [here](https://github.com/JetBrains/compose-hot-reload/pull/449); note: be careful to not delete the `teamcity` branch after PR is merged)
* `teamcity` branch contains some [leftover commits](https://github.com/JetBrains/compose-hot-reload/commits/teamcity/) after your experiments, they need to be cleaned up. In general I would recommend using the Kotlin DSL for changing the build configs, manual edits via UI can lead to unnecessary conflicts. It is ok to use UI for experiments, but only if you are sure that all the commits with changes can then be fully removed

PR #530, opened 2026-05-28
[mcpserver] Add multi-window support

pjBooms:
### Summary

MCP window-targeting tools (`take_screenshot`, `get_semantic_tree`, `click`, `long_click`, `type_text`, `scroll`, `scroll_to_index`) were ambiguous with multiple Compose windows — they broadcast and collected whichever response arrived first, so AI agents had no way to pick a specific window.

This PR:

1. Adds a `list_windows` MCP tool so agents can discover registered windows.
2. Adds an optional `window_id` parameter to every window-targeting tool.
3. Changes the default (no `window_id`) to operate on the first registered window instead of broadcasting.
4. Adds a `title` field to `WindowsState.WindowState` so agents can differentiate windows by their on-screen title.

### Protocol changes

- `ScreenshotRequest`, `SemanticTreeRequest`, `UIActionRequest` now carry an optional `windowId: WindowId?` field. Encoders updated to round-trip the field using the same nullable pattern already used by `UIRendered`.
- Application-side handlers in `DevelopmentEntryPoint` filter requests: `request.windowId == null || request.windowId == windowId`. Unknown window IDs are simply dropped on the application side.
- `WindowsState.WindowState` gains a `title: String` field, populated from `java.awt.Frame.title` / `Dialog.title` and refreshed whenever the window title changes.
- Wire compatibility: an older application talking to a newer MCP server decodes `windowId = null` and falls back to the legacy broadcast behaviour. Acceptable for a dev tool.

### MCP server

- New `list_windows` tool returns a JSON array of `{id, title, x, y, width, height}` per window, in `WindowsState` insertion order. The `title` lets agents identify windows by what the user actually sees, not just opaque IDs.
- Resolution: explicit `window_id` is validated against `WindowsState.windows`; unknown IDs return `Window '<id>' not found`. Without a `window_id`, the server resolves to the first registered window before sending.

### Tests

| Layer | What was added |
|---|---|
| **Serialization** (`hot-reload-orchestration`) | 16 new tests in `SerializationTest.kt`: round-trip for `ScreenshotRequest`/`SemanticTreeRequest` (null + windowId), `ScreenshotResult`, `SemanticTreeResult`, all five `UIAction` subclasses via `UIActionRequest`, equality test (`windowId` differentiates), and `UIActionResult` success/failure. These six messages had **never** been covered — the gap predates this branch and was inherited from the original MCP commit. |
| **WindowsState** (`hot-reload-devtools-api`) | `WindowsStateTest` extended to round-trip the new `title` field. |
| **MCP server unit** (`hot-reload-mcp`) | 11 new tests in `McpServerTest.kt` + an `awaitWindows` helper: `list_windows` (disconnected / empty / insertion-ordered JSON including titles / live title updates); for `take_screenshot`, `get_semantic_tree`, `click` — explicit `window_id` is forwarded on the wire, default resolves to first window, unknown id returns `Window '<id>' not found`. |
| **Integration** (`tests/`) | 3 new non-headless tests in `MultiWindowScreenshotIntegrationTest.kt` driving a real two-window Compose app (red + blue): explicit `windowId` returns that window's pixels (centre-pixel channel check), `null` broadcasts to both windows (pins wire-level contract), unknown `windowId` is dropped (no response within 3 s). |

### DIFF
--- a/hot-reload-devtools-api/src/main/kotlin/org/jetbrains/compose/devtools/api/WindowsState.kt
+++ b/hot-reload-devtools-api/src/main/kotlin/org/jetbrains/compose/devtools/api/WindowsState.kt
@@ -40,12 +40,13 @@ public class WindowsState(
         return "WindowsState($windows)"
     }
 
-    public class WindowState(
+    public class WindowState @JvmOverloads constructor(
         public val x: Int,
         public val y: Int,
         public val width: Int,
         public val height: Int,
-        public val isAlwaysOnTop: Boolean
+        public val isAlwaysOnTop: Boolean,
+        public val title: String? = null,
     ) {
         private val hashCode: Int = run {
             var result = x.hashCode()
@@ -53,6 +54,7 @@ public class WindowsState(
             result = 31 * result + width
             result = 31 * result + height
             result = 31 * result + isAlwaysOnTop.hashCode()
+            result = 31 * result + (title?.hashCode() ?: 0)
             result
         }
 
@@ -69,11 +71,12 @@ public class WindowsState(
             if (this.width != other.width) return false
             if (this.height != other.height) return false
             if (this.isAlwaysOnTop != other.isAlwaysOnTop) return false
+            if (this.title != other.title) return false
             return true
         }
 
         override fun toString(): String {
-            return "WindowState(x=$x, y=$y, width=$width, height=$height, isAlwaysOnTop=$isAlwaysOnTop)"
+            return "WindowState(x=$x, y=$y, width=$width, height=$height, isAlwaysOnTop=$isAlwaysOnTop, title=$title)"
         }
     }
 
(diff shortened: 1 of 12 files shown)


hot-reload-devtools-api/src/main/kotlin/org/jetbrains/compose/devtools/api/WindowsState.kt
```diff
@@ -40,19 +40,21 @@ public class WindowsState(
...
 
-    public class WindowState(
+    public class WindowState @JvmOverloads constructor(
         public val x: Int,
         public val y: Int,
         public val width: Int,
         public val height: Int,
-        public val isAlwaysOnTop: Boolean
+        public val isAlwaysOnTop: Boolean,
+        public val title: String = "",
```
AbdullinAM: Why not make it `null` by default? Thats what I would expect if there is no title available

AbdullinAM (approved):

hot-reload-devtools-api/src/main/kotlin/org/jetbrains/compose/devtools/api/WindowsState.kt
pjBooms: Makes sense. Done.

PR #531, opened 2026-05-29
Show error details in the notification overlay

AbdullinAM:


vanniktech (commented):
I'm excited! If there's a build with this, I'd like to test it.

AbdullinAM:
@vanniktech There is no build yet. I can share how it looks in the current version:

Screenshot 2026-05-29 at 14 13 02

vanniktech:
This looks gorgeous. The copy button will only copy the file path (including line number + line colon) such that I can instantly jump to IntelliJ and fix it right?

AbdullinAM:
Unfortunately, no. Right now it will copy the whole error message, even the parts that don't fit in the notification pop up: even in this example the full message is 5+ lines, we can only fit 3 on the screen.

Maybe we can add a second button that will detect file likns in the message and copy only them, I'm not sure if we can display them properly. @pjBooms wdyt about that idea?

hot-reload-devtools-api/api/hot-reload-devtools-api.api
```diff
@@ -138,9 +138,10 @@ public abstract class org/jetbrains/compose/devtools/api/ReloadState : org/jetbr
 
 public final class org/jetbrains/compose/devtools/api/ReloadState$Failed : org/jetbrains/compose/devtools/api/ReloadState {
 	public static final field $stable I
-	public fun <init> (Lkotlin/time/Instant;Ljava/lang/String;)V
```
pjBooms, line 141: It seems to be backward incompatible change. Should be fixed with @JvmOverloads

vanniktech:
For me the most important would be that the copy button just copies the full file link so everything after `file://` up to `:99:2` from the screenshot.

Edit: If there are more than one file, just the first one is enough. Then I'd recompose and go to the next file.

AbdullinAM:
@vanniktech prototype:
Screenshot 2026-05-29 at 21 44 28
Screenshot 2026-05-29 at 21 44 20

hot-reload-runtime-jvm/src/main/kotlinUI/org/jetbrains/compose/hotReloadUI/widgets/CopyToClipboard.kt
```diff
@@ -0,0 +1,89 @@
...
+    var copyAll by remember { mutableStateOf(false) }
+    LaunchedEffect(copyAll) {
+        if (copyAll) {
+            clipboard.setClipEntry(ClipEntry(StringSelection(content())))
+            copyAll = false
+        }
+    }
+
+    val shape = RoundedCornerShape(4.dp)
+    TooltipArea(
```
AbdullinAM, line 60: My logic is: `TooltipArea` is experimental, but it is quite stable in Compose Desktop. WDYT?

If necessary, I can implement custom tooltips similar to how it is done in devtools. In that case, I would also suggest we create a separate module to share UI components between devtools and reload overlays. We have re-implemented some stuff already

vanniktech:
I can't wait to test this out!

hot-reload-devtools-api/src/main/kotlin/org/jetbrains/compose/devtools/api/ReloadState.kt
```diff
@@ -133,6 +141,7 @@ internal class ReloadStateEncoder : OrchestrationStateEncoder<ReloadState> {
             writeFields(
                 "time" to encodeByteArray { writeLong(state.time.toEpochMilliseconds()) },
                 "reason" to state.reason.encodeToByteArray(),
+                "details" to state.details?.joinToString("\n")?.encodeToByteArray()
```
pjBooms: it seems emptyList() becomes listOf("") after decoding.

hot-reload-runtime-jvm/src/main/kotlinUI/org/jetbrains/compose/hotReloadUI/ErrorNotificationOverlayEffect.kt
```diff
@@ -165,15 +172,51 @@ internal fun ErrorNotification(
...
+                    color = ColorProducer { Color.White }
+                )
+
+                if (details.isNotBlank()) {
+                    Spacer(Modifier.width(4.dp))
+                    Spacer(Modifier.weight(1f))
+                    if (filePath.isNotBlank()) {
+                        CopyToClipboardButton(
+                            icon = fileSvg,
+                            tooltip = "Copy path to the file",
```
pjBooms, line 195: Will it copy the path including "file://", line and pos?
It seems the request was only about the file path only.  Though I am not sure actually the usage scenario here. @vanniktech can you elaborate?

hot-reload-runtime-jvm/src/main/kotlinUI/org/jetbrains/compose/hotReloadUI/widgets/CopyToClipboard.kt
pjBooms: I have no objections for using it

hot-reload-runtime-jvm/src/main/kotlinUI/org/jetbrains/compose/hotReloadUI/ErrorNotificationOverlayEffect.kt
AbdullinAM: Yes, that makes sense. I removed the "file://" prefix

pjBooms: What about line and col? Are they needed?

pjBooms (approved):

hot-reload-runtime-jvm/src/main/kotlinUI/org/jetbrains/compose/hotReloadUI/ErrorNotificationOverlayEffect.kt
AbdullinAM: I would say yes, they allow you to go to the specific place in the file (presumably where the error is), instead of just opening it at the top

vanniktech: @pjBooms if the error from the kotlin compliler is:

> e: file:///Users/niklas/dev/GitHub/vanniktech/feature-flashcards/src/commonMain/kotlin/com/vanniktech/feature/flashcards/preferences/FlashcardsImportExport.kt:165:106 Unresolved reference 'it'.


then I want the file path with line and colon:

> /Users/niklas/dev/GitHub/vanniktech/feature-flashcards/src/commonMain/kotlin/com/vanniktech/feature/flashcards/preferences/FlashcardsImportExport.kt:165:106

Such that I can do:

Screenshot 2026-06-03 at 11 10 55

PR #533, opened 2026-06-04
[mcpserver] Startup failure fixes

pjBooms:
[CMP-10258](https://youtrack.jetbrains.com/issue/CMP-10258) fixes: 

 - The MCP server fails to start when there is no pid directory
 - Redirect System.out to stderr for MCP server to keep stdout for JSON-RPC transport

### DIFF
--- a/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
+++ b/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
@@ -39,6 +39,10 @@ private val logger = createLogger()
 
 @OptIn(InternalHotReloadApi::class)
 fun main(args: Array<String>) {
+    // The MCP stdio transport reserves stdout for the JSON-RPC protocol.
+    val protocolOut = System.out
+    System.setOut(System.err)
+
     val pidFile = HotReloadEnvironment.pidFile
         ?: args.firstOrNull()?.let { Path.of(it) }
         ?: run {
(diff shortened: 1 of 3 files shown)


hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpMain.kt
```diff
@@ -38,6 +39,10 @@ private val logger = createLogger()
 
 @OptIn(InternalHotReloadApi::class)
 fun main(args: Array<String>) {
+    // The MCP stdio transport reserves stdout for the JSON-RPC protocol.
+    val protocolOut = System.out
```
AbdullinAM, line 43: If we encounter conflicts with the std output streams, maybe its not worth using them as transport at all? Does MCP Kotlin SDK provide ways to use something different?

Redirecting input streams introduces a lot of ways to easily break something and just in general makes the behaviour of the program non obvious

pjBooms: Alternative is HTTP/SSE. It works if you have something permanent. Like IDEA starts and always launches its MCP server on a pre-allocated port that an agent may use. We are in different situation: agents start the mcp server because noone else can start it. And this scenario usually means JSON-RPC over stdout, afaik.

pjBooms: Redirecting stdout to stderr is quite common for MCP servers, afaik

AbdullinAM (approved):

PR #536, opened 2026-06-05
[mcpserver] Add `reload` tool to trigger hot reload

pjBooms:
Adds a `reload` tool to the MCP server so an AI agent can apply source
edits to a running Compose app without restarting it.

### DIFF
--- a/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
+++ b/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
@@ -20,6 +20,8 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.flow.filterIsInstance
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.withTimeoutOrNull
 import kotlinx.io.asSink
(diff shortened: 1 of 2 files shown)


hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -259,6 +288,103 @@ private suspend fun handleStatus(orchestration: StateFlow<OrchestrationHandle?>)
...
+            isError = true
+        ).also { logger.info { "reload: no application connected" } }
+
+    val timeoutSeconds = (toolRequest.arguments?.get("timeout_seconds")?.jsonPrimitive?.intOrNull
+        ?: DEFAULT_RELOAD_TIMEOUT_SECONDS).coerceAtLeast(1)
+
+    val request = RecompileRequest()
+    logger.info { "reload: sending RecompileRequest '${request.messageId}' (timeout ${timeoutSeconds}s)" }
+
+    /*
```
AbdullinAM, line 312: Maybe we can reuse/adapt `OrchestrationHandle.requestResponse` here somehow? Just because we now have two separate palces where we have to explain hot channels work

AbdullinAM (approved):

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: Valid. Thanks. Addressed

AbdullinAM (approved):

PR #537, opened 2026-06-05
[mcpserver] Add resize_window tool to the mcp server

pjBooms:


hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -560,6 +576,68 @@ private suspend fun handleUIAction(
...
+        )
+    val height = args["height"]?.jsonPrimitive?.intOrNull
+        ?: return CallToolResult(
+            content = listOf(TextContent("Missing required parameter 'height' (integer)")),
+            isError = true
+        )
+
+    val resolvedId = try {
+        resolveWindowId(handle, args)
+    } catch (e: IllegalArgumentException) {
```
AbdullinAM: Maybe we can provide a proper custom exception here? In that case, you can just combine the two `try/catch`es and just have one

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -560,6 +576,68 @@ private suspend fun handleUIAction(
...
+    }
+
+    return try {
+        val resizeRequest = WindowResizeRequest(width = width, height = height, windowId = resolvedId)
+        logger.info { "resize_window: sending request '${resizeRequest.messageId}' ${width}x${height}" }
+        val result = handle.requestResponse<WindowResizeResult>(resizeRequest) {
+            it.windowResizeRequestId == resizeRequest.messageId
+        }
+
+        if (result == null) {
```
AbdullinAM: Why not combine this into a single `if/else` or `when` expression?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -560,6 +576,68 @@ private suspend fun handleUIAction(
...
+            logger.warn("resize_window: timed out waiting for response")
```
AbdullinAM: I think it will be more clear to just log the `CallToolResult` one time at the end of the function, instead of writing a separate `log` call before each `return` that basically duplicates the message

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationProtocolVersion.kt
```diff
@@ -82,5 +82,7 @@ internal val Class<out OrchestrationMessage>.availableSinceVersion: Orchestratio
         OrchestrationMessage.SemanticTreeResult::class.java -> OrchestrationVersion.v1_5
         OrchestrationMessage.UIActionRequest::class.java -> OrchestrationVersion.v1_5
         OrchestrationMessage.UIActionResult::class.java -> OrchestrationVersion.v1_5
+        OrchestrationMessage.WindowResizeRequest::class.java -> OrchestrationVersion.v1_5
```
AbdullinAM: I don't remember: were the previously declared orchestration messages already released? If so, I thin it makes sense to declare these new messages as `v1_6`

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/DevelopmentEntryPoint.kt
```diff
@@ -101,6 +102,15 @@ public fun DevelopmentEntryPoint(
                     handleUIActionRequest(request, window).sendAsync()
                 }
         }
+
+        LaunchedEffect(Unit) {
+            orchestration.asFlow()
+                .filterIsInstance<WindowResizeRequest>()
+                .filter { request -> request.windowId == null || request.windowId == windowId }
```
AbdullinAM: Would this mean that in case `request.windowId == null` we resize all the windows? And according to the [docs](https://github.com/JetBrains/compose-hot-reload/tree/staging#targeting-a-specific-window) --- it should only work for the "first" window? I think it also applies to all other requests

AbdullinAM: Also, the number of very similar
```
        LaunchedEffect(Unit) {
            orchestration.asFlow()
                .filterIsInstance<RequestName>()
                .filter { request -> request.windowId == null || request.windowId == windowId }
                .collect { request ->
                    handleRequestName(request, window).sendAsync()
                }
        }
```

code blocks in `DevelopmentEntryPoint` continues growing. I would at least consider creating a separate utility function to simplify it and reduce duplication

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/windowResizeHandler.kt
```diff
@@ -0,0 +1,40 @@
...
+            isSuccess = false,
+            errorMessage = "Invalid window size ${request.width}x${request.height}: width and height must be positive",
+        )
+    }
+
+    return try {
+        window.setSize(request.width, request.height)
+        window.validate()
+        WindowResizeResult(windowResizeRequestId = request.messageId, isSuccess = true)
+    } catch (e: Exception) {
```
AbdullinAM: What exceptions are we expecting here? According to the docs, neither `setSize` nor `validate` throw any exceptions

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/DevelopmentEntryPoint.kt
pjBooms: > Would this mean that in case `request.windowId == null` we resize all the windows? And according to the [docs](https://github.com/JetBrains/compose-hot-reload/tree/staging#targeting-a-specific-window) --- it should only work for the "first" window? I think it also applies to all other requests

MCP never sends `null` here, so the doc is still correct. I did not want to apply the filtering logic in the runtime.

AbdullinAM: But still, it means that either docs or implementation is incorrect. If we declare that `windowId` can be omitted, why is it not supported by the MCP server? 

Also, it means that the semantics of the `windowId` are not declared anywhere: the orchestration message [docs](https://github.com/JetBrains/compose-hot-reload/blob/c104eee9a636c54de7397cb022fd7892594c6710/hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt#L601) do not explain it, and the MCP server docs are incorrect.

I really don't like this situation. I think the behaviour should be explicitly described. Its even fine if the semantics of the orchestration messages are different from the semantics of the MCP tools, but it should be easily detectable. Right now I have to read 2-3 files to understand how and why this works. 

And I think it would be a good idea to add docs to the orchestration messages in any case, even if it would just duplicate mcp server docs. We have to remember that the orchestration protocol can be potentially used by external tools, so it is good to have clear semantics set

pjBooms: > But still, it means that either docs or implementation is incorrect. 

Both correct. The MCP server docs describe its actual behavior. 
Orchestration messages docs were not complete as they did not describe broadcasting semantic of null `window_id`.  Fixed, please check.

However, I think that broadcasting semantic for newly introduced `WindowResizeRequest`  is a bit weird (unlike other requests that can be useful for some scenarios), so I have made `window_id` required for it.

AbdullinAM: Do you think broadcasting all the other requests makes sense in that case? If you now broadcast the semantic tree request to all windows, you will get back `N` semantic tree results and you will not be able to tell which window each of them belongs to. None of the responses contain a `windowId` to identify the result.

In my view, the resize request is not too different from all other requests. I don't see a big difference between resizing all windows and taking screenshot of all windows.

In that sense, I think it either (1) makes sense to require window id for each one of the requests or (2) to make window resize broadcastable too and add `windowId` to all the responses. I personally would vote for option (1), but its up to you

AbdullinAM: Other fixes are good, thanks

pjBooms: >None of the responses contain a windowId to identify the result.

We can add optional `window_id` to other responses but changing size to identical for all windows is really strange. So I would keep this asymmetry. WDYT?

AbdullinAM (approved):

PR #538, opened 2026-06-08
[mcpserver] Support Dialog/ModalBottomSheet/Popup roots in semantic tree

pjBooms:
Fixes: [CMP-10282](https://youtrack.jetbrains.com/issue/CMP-10282)

### DIFF
--- a/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/runHeadless.kt
+++ b/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/runHeadless.kt
@@ -263,21 +263,17 @@ public suspend fun runHeadlessApplication(
             if (message is OrchestrationMessage.SemanticTreeRequest) {
                 logger.info("Capturing semantic tree: '${message.messageId}'")
                 scene.render(virtualTime)
-                val rootNode = scene.tryGetRootSemanticsNode()
-                val json = if (rootNode != null) buildSemanticTreeJson(rootNode)
-                else """{"error":"No semantic owners available"}"""
                 orchestration.send(
                     OrchestrationMessage.SemanticTreeResult(
                         semanticTreeRequestId = message.messageId,
-                        tree = json,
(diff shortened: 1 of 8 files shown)


AbdullinAM (approved):

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/runHeadless.kt
```diff
@@ -349,19 +345,48 @@ private fun issueSilenceWarning(silence: Duration, silenceTimeout: SilenceTimeou
...
 /**
- * Extracts the root [SemanticsNode] from an [ImageComposeScene] using reflection.
+ * Extracts every root [SemanticsNode] from an [ImageComposeScene] using reflection.
+ *
+ * The scene exposes a main owner plus one owner per attached layer (Dialog / ModalBottomSheet /
+ * Popup), each carrying its own semantics root. We collect all of them.
  */
-private fun ImageComposeScene.tryGetRootSemanticsNode(): SemanticsNode? {
-    // ImageComposeScene.scene.mainOwner.semanticsOwner
+private fun ImageComposeScene.tryGetAllRootSemanticsNodes(): List<SemanticsNode> {
```
MatkovIvan, line 353: Why do you do this with reflection?
It's exposed via public API - `ImageComposeScene.semanticsOwners`

pjBooms: It is introduced in 1.10 afaik. We still support and compile against Compose 1.8.2.

MatkovIvan: You're relying on a private API shape that may (and will) be changed without notice. This approach does not improve compatibility, but rather makes it worse

pjBooms: Ok, please take a look here -- https://github.com/JetBrains/compose-hot-reload/pull/539

PR #539, opened 2026-06-08
[mcpserver] Use `getSemanticsOwners()` public API for extracting semantic tree roots

pjBooms:


MatkovIvan (approved):

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/runHeadless.kt
```diff
@@ -345,12 +346,27 @@ private fun issueSilenceWarning(silence: Duration, silenceTimeout: SilenceTimeou
...
 /**
- * Extracts every root [SemanticsNode] from an [ImageComposeScene] using reflection.
+ * Extracts every root [SemanticsNode] from an [ImageComposeScene]: a main owner plus one owner per
+ * attached layer (Dialog / ModalBottomSheet / Popup), each carrying its own semantics root.
  *
- * The scene exposes a main owner plus one owner per attached layer (Dialog / ModalBottomSheet /
- * Popup), each carrying its own semantics root. We collect all of them.
+ * Prefers the public `ImageComposeScene.getSemanticsOwners()`. That accessor only exists since
+ * Compose 1.10, while this module is compiled against the minimum supported Compose (1.8.2), so it
+ * is invoked reflectively; on older Compose we fall back to reading the scene owners directly.
```
MatkovIvan: I guess we can just bump minimal compose to 1.10 and drop all reflection hacks here

pjBooms: > I guess we can just bump minimal compose to 1.10 and drop all reflection hacks here

imho it is too fast, though bumping to 1.9.0 seems to be ok.
@kropp , @AbdullinAM WDYT?

AbdullinAM: Yes, I think updating minimal version to 1.9.0 is ok

MatkovIvan: Updating it to 1.9.0 doesn't solve the problem that we need API from 1.10.

> imho it is too fast

Why? Currently we're working on 1.12, so it will still support two last stable versions (with 1.12 release 3 last stable).
Also, previous versions of Hot Reload will be still there. Since it's an independent "library"(plugin), it's not about breaking anyone but preventing updating Hot Reload independently from Compose.
I doubt that our QAs have the capacity for testing new versions of Hot Reload against old version of Compose anyway.

I don't think that it justifies using reflection/private APIs here

sekater: I would support Ivan's opinion, looks like the current 1.12 version and the two stable previous versions are sufficient

kropp: Agree with @MatkovIvan

pjBooms: Ok, dropped 1.8.2 and 1.9.0 in https://github.com/JetBrains/compose-hot-reload/pull/540

PR #540, opened 2026-06-08
[tests] Drop 1.8.2 and 1.9.0 Compose versions

pjBooms:


dependencies.toml
```diff
@@ -1,7 +1,8 @@
 [versions]
 kotlin = "2.3.21"
 compose = "1.11.0"
-composeMin = { prefer = "1.8.2" }
+composeMin = { prefer = "1.10.0" }
+composeMaterial3 = "1.10.0-alpha05"
```
AbdullinAM, line 5: Why do we need this?

hot-reload-runtime-jvm/src/main/kotlinUI/org/jetbrains/compose/hotReloadUI/widgets/CopyToClipboard.kt
```diff
@@ -39,7 +40,7 @@ import org.jetbrains.compose.hot_reload.hot_reload_runtime_jvm.generated.resourc
 import org.jetbrains.compose.resources.painterResource
 import java.awt.datatransfer.StringSelection
 
-@OptIn(ExperimentalFoundationApi::class)
+@OptIn(ExperimentalFoundationApi::class, ExperimentalComposeUiApi::class)
```
AbdullinAM, line 43: Why did the opt-ins extend?

AbdullinAM (approved):
I'm still not really a fan of updating the minimal versions, but if the new features require it --- so be it. HOpefully we will be able to collect enough feedback during alpha-beta releases to ensure the our users will not be affected by the changes

dependencies.toml
pjBooms: Material3 versioning was decoupled from Compose since 1.10.0 because it is based on 1.5.0 alphas (not stable). See https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.10.0

hot-reload-runtime-jvm/src/main/kotlinUI/org/jetbrains/compose/hotReloadUI/widgets/CopyToClipboard.kt
pjBooms: https://github.com/JetBrains/compose-multiplatform-core/pull/2538

dependencies.toml
MatkovIvan: We can still rely on the latest stable in the case of material3, in general there is no requirement to use "1.10"/same as base CMP

Please make sure that there will be no situation where "stable" CHR depends on "alpha" library.
At least in prod, for tests it's ok

PR #541, opened 2026-06-09
[mcpserver] Add windowId to ScreenshotResult/SemanticTreeResult/UIActionResult

pjBooms:


hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
```diff
@@ -329,13 +329,17 @@ internal constructor() : OrchestrationPackage(), Serializable {
      * @param data the raw image data, empty if [isSuccess] is false
      * @param isSuccess whether the screenshot was captured successfully
      * @param errorMessage an error description if [isSuccess] is false
+     * @param windowId the window that produced this result, or `null` when produced without a window
+     * (for example, in headless mode). Lets a broadcast [ScreenshotRequest] (with a `null` windowId)
```
AbdullinAM: > Lets a broadcast [ScreenshotRequest] (with a `null` windowId) be attributed to the window each response came from

I would rephrase that sentence, its meaning is not very clear

AbdullinAM: This applies to every changed message

AbdullinAM (approved):

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
```diff
@@ -329,13 +329,17 @@ internal constructor() : OrchestrationPackage(), Serializable {
...
+     * be attributed to the window each response came from.
      */
-    public class ScreenshotResult(
+    public class ScreenshotResult @JvmOverloads constructor(
```
AbdullinAM, line 338: Didn't we also want to add an experimental annotation?

pjBooms: It is in another PR. Can do in one

AbdullinAM: I don't see a reason to split it into two PRs, but up to you

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
pjBooms: Rephrased, plaese take a look

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
pjBooms: Ok, added here

AbdullinAM: I see. This version is better, although I would argue that we don't even need this sentence at all. But again, this is more personal preference than anything

AbdullinAM (approved):

hot-reload-orchestration/src/main/kotlin/org/jetbrains/compose/reload/orchestration/OrchestrationMessage.kt
AbdullinAM: This comment was meant for the thread about the docs

pjBooms: Ok, let's be short :)

PR #549, opened 2026-06-16
[mcpserver] Add get_logs tool

pjBooms:
**Implementation note:** retrieves (limited) logs from `<pidFile>.chr.log` that allows an AI agent to see the picture from very start of the application.

### DIFF
--- a/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/logging.kt
+++ b/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/logging.kt
@@ -9,6 +9,7 @@ import org.jetbrains.compose.reload.core.HotReloadEnvironment
 import org.jetbrains.compose.reload.core.Logger
 import org.jetbrains.compose.reload.core.Queue
 import org.jetbrains.compose.reload.core.WorkerThread
+import org.jetbrains.compose.reload.core.chrLogFile
 import org.jetbrains.compose.reload.core.createLogger
 import org.jetbrains.compose.reload.core.debug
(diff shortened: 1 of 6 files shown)


AbdullinAM (commented):
What is the reasoning for accessing the file instead of receiving logs via orchestration? So that we can see logs that are collected before the MCP server start? Do you expect such use case to be common? 

We are also potentially opening ourself to possibility of parallel file write/read conflicts, although in this case I think it is not too critical.

pjBooms:
> What is the reasoning for accessing the file instead of receiving logs via orchestration? So that we can see logs that are collected before the MCP server start? Do you expect such use case to be common?

I can easily imagine situation when a user get a problem with the UI and then open an AI agent asking: "what happened?". If the agent connects to the app only this time it won't have any clue in case of receiving logs via orchestrartion.

I also thought that we could ask dev tools for the logs but they are limited there and it seems that transferring them won't be significantly better than reading the file. 

Third alternative was introducing global state with the logs  but it seems to be bad idea as it brings quadratic time to update the state with any new log entry.

>We are also potentially opening ourself to possibility of parallel file write/read conflicts, although in this case I think it is not too critical.

I believe agents will use the tool  rarely, when `status` and `get_ui_error` tools won't give them enough information.
Moreover we may change the implementation if we encounter problems with the approach.

AbdullinAM:
> Third alternative was introducing global state with the logs but it seems to be bad idea as it brings quadratic time to update the state with any new log entry.

Is also linearly increases the size of the orchestration states (that you need to constantly transfer over network), so this is a really bad idea

AbdullinAM (approved):

PR #554, opened 2026-06-18
[mcpserver] Introduce `await_reload` tool for  `--auto` build mode

pjBooms:
Waits for the reload to finish. Asks the client to poll the status on timeout. 
May/should be called for `--auto` mode.

### DIFF
--- a/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
+++ b/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
@@ -18,6 +18,7 @@ import io.modelcontextprotocol.kotlin.sdk.types.TextContent
 import io.modelcontextprotocol.kotlin.sdk.types.ToolSchema
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.flow.filterIsInstance
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.mapNotNull
@@ -37,6 +38,7 @@ import kotlinx.serialization.json.intOrNull
 import kotlinx.serialization.json.jsonPrimitive
 import kotlinx.serialization.json.put
 import kotlinx.serialization.json.putJsonObject
+import org.jetbrains.compose.devtools.api.BuildModeState
 import org.jetbrains.compose.devtools.api.ReloadCountState
 import org.jetbrains.compose.devtools.api.ReloadState
 import org.jetbrains.compose.devtools.api.UIErrorState
@@ -44,6 +46,7 @@ import org.jetbrains.compose.devtools.api.WindowsState
 import org.jetbrains.compose.reload.DelicateHotReloadApi
 import org.jetbrains.compose.reload.core.HOT_RELOAD_VERSION
 import org.jetbrains.compose.reload.core.WindowId
+import org.jetbrains.compose.reload.core.chrLogFile
 import org.jetbrains.compose.reload.core.createLogger
 import org.jetbrains.compose.reload.core.debug
 import org.jetbrains.compose.reload.core.info
@@ -67,6 +70,7 @@ import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.RestartRe
 import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.WindowResizeRequest
 import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.WindowResizeResult
 import org.jetbrains.compose.reload.orchestration.asChannel
+import org.jetbrains.compose.reload.orchestration.flowOf
 import java.io.OutputStream
 import java.nio.file.Path
 import java.util.Base64
@@ -75,6 +79,7 @@ import kotlin.io.path.createParentDirectories
 import kotlin.io.path.isRegularFile
 import kotlin.io.path.writeBytes
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
 
 private val logger = createLogger()
@@ -82,13 +87,13 @@ private val logger = createLogger()
 internal suspend fun startMcpServer(
     orchestration: StateFlow<OrchestrationHandle?>,
     protocolOut: OutputStream,
-    logFile: Path,
+    pidFile: Path,
 ) {
     val transport = StdioServerTransport(
         inputStream = System.`in`.asSource().buffered(),
         outputStream = protocolOut.asSink().buffered()
     )
-    startMcpServer(orchestration, transport, logFile)
+    startMcpServer(orchestration, transport, pidFile)
 }
 
 /** A single input property in a tool's JSON schema. */
@@ -142,7 +147,7 @@ private fun toolSchema(
 internal suspend fun startMcpServer(
     orchestration: StateFlow<OrchestrationHandle?>,
     transport: Transport,
-    logFile: Path,
+    pidFile: Path,
 ) {
     val server = Server(
         serverInfo = Implementation(name = "compose-hot-reload", version = HOT_RELOAD_VERSION),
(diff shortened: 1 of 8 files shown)


hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/PidFileInfo.kt
```diff
@@ -35,14 +35,25 @@ public fun PidFileInfo(binary: ByteArray): Try<PidFileInfo> = Try {
...
-public data class PidFileInfo(val pid: Long?, val orchestrationPort: Int?) {
+/**
+ * @param pid The process ID of the running application.
+ * @param orchestrationPort The TCP port on which the orchestration server is listening.
+ * @param buildContinuous `true` when the application runs with continuous build / auto-reload (the `--auto` flag).
+ */
+public data class PidFileInfo @JvmOverloads constructor(
+    val pid: Long?,
+    val orchestrationPort: Int?,
+    val buildContinuous: Boolean = false,
```
AbdullinAM: It feels like we're adding this property here just because we can, not because it fits here well. Is there any other way to provide this information?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -560,6 +594,116 @@ private suspend fun handleListWindows(orchestration: StateFlow<OrchestrationHand
         textResult(json.toString())
     }
 
+/**
+ * How long to wait for the continuous build to pick up the latest edit (i.e. for [ReloadState] to
+ * become [ReloadState.Reloading]) before concluding that nothing changed.
+ */
+private val WAIT_FOR_RELOAD_TO_START_TIMEOUT = 1500.milliseconds
```
AbdullinAM: Maybe we can extract all these constants into a separate config object/file? Because right now they are scattered across the file, its hard to find them and they make the already big file even bigger

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -560,6 +594,116 @@ private suspend fun handleListWindows(orchestration: StateFlow<OrchestrationHand
...
+    data class ReloadingFinished(val result: ReloadResult) : WaitForReloadResult
+}
+
+/**
+ * Continuous ('--auto') variant of [handleReload]: does not send a [RecompileRequest]; instead observes
+ * the autonomous reload via [ReloadCountState]/[ReloadState] (both broadcast in either build mode) and
+ * reports its result. Returns within [timeoutSeconds], degrading to the same 'poll status' response as
+ * the explicit path if the reload is still running.
+ */
+private suspend fun handleReloadContinuous(
```
AbdullinAM: Two notes that I have:
* We have a potential problem on the recompiler level: if the reload request comes in the auto mode, the reload state hangs indefinitely. I think it should be addressed in the recompiler itself
* To simplify the logic, maybe its worth making this a separate tool (e.g. "await reload"). We can explain to the agent that the reload tool works in manual mode and await reload works in the auto mode only. Then it can select the proper tool itself. And the incorrect reload request will be handled by the recompiler, the incorrect await will just end in a timeout

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -560,6 +594,116 @@ private suspend fun handleListWindows(orchestration: StateFlow<OrchestrationHand
...
+private val WAIT_FOR_RELOAD_TO_START_TIMEOUT = 1500.milliseconds
+
+/**
+ * After a continuous build finishes, the reload counters still can be not updated.
+ * We may wait a bit to decide if nothing changed.
+ */
+private val WAIT_FOR_COUNTERS_UPDATE_TIMEOUT = 1.seconds
+
+/** Result of observing an autonomous (continuous-mode) reload. */
+private sealed interface ReloadResult {
```
AbdullinAM: Tbh I don't fully see why do we need these separate classes instead of just using `ReloadState`. What information is it lacking?

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -560,6 +594,116 @@ private suspend fun handleListWindows(orchestration: StateFlow<OrchestrationHand
...
+        (handle.states.get(ReloadState).value as? ReloadState.Failed)?.reason
+
+    // A reload completed (relative to [baseline]) iff one of the counters advanced.
+    suspend fun reloadResultByCountersAdvance(current: ReloadCountState): ReloadResult? = when {
+        current.failedReloads > baseline.failedReloads -> ReloadResult.Failed(failureReason())
+        current.successfulReloads > baseline.successfulReloads -> ReloadResult.Reloaded
+        else -> null
+    }
+
+    val countFlow = handle.states.flowOf(ReloadCountState)
```
AbdullinAM: I really think that this logic is overcomplicated. I would opt to just stick with one state and use it as a reference point.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: > I think it should be addressed in the recompiler itself

Addressed here: https://github.com/JetBrains/compose-hot-reload/pull/557

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/PidFileInfo.kt
pjBooms: `buildContinuous` is a runtime _mode_, not application state. It's set once at startup and never changes. So it's a static fact about the process, not something that changes during execution. 

Alternative is to ask the process about its mode. It requires introducing new `Request/Result` messages. Just more code, imho.

Introducing a global state is not a way, imho, because it is not actually state, it never updates.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: It's state machine classes served just for the await reload implementation. While ReloadState is global state. We are waiting the state changes and having results of that waits

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: It is related to a concrete logic, I usually keep such private constants close to it just because its easer to find, imho

hot-reload-core/src/main/kotlin/org/jetbrains/compose/reload/core/PidFileInfo.kt
pjBooms: After introducing separate tool `await_reload`, decided to move the flag to experimental `BuildModeState`.
We may return back to pid file implementation, if you think it is better.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: >To simplify the logic, maybe its worth making this a separate tool (e.g. "await reload").

Re-implemented this way in this PR.

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: Finally, just use `ReloadState` without obsersving reload counts

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: Ok, followed this way (one state) in current version.

PR #555, opened 2026-06-18
[mcpserver] CMP-10329: Add save_to parameter to take_screenshot tool

pjBooms:
Allows the tool to save screenshot image bytes to disk.

Fixes: [CMP-10329](https://youtrack.jetbrains.com/issue/CMP-10329)

### DIFF
--- a/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
+++ b/hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
@@ -71,7 +71,9 @@ import java.io.OutputStream
 import java.nio.file.Path
 import java.util.Base64
 import kotlin.io.path.bufferedReader
+import kotlin.io.path.createParentDirectories
 import kotlin.io.path.isRegularFile
+import kotlin.io.path.writeBytes
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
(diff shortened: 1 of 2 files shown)


hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
```diff
@@ -793,8 +800,24 @@ private suspend fun handleTakeScreenshot(
             }
             else -> {
                 logger.debug { "take_screenshot: received ${screenshot.data.size} bytes (${screenshot.format})" }
+                val saveMessage: String? = if (saveTo != null) {
+                    try {
+                        saveTo.createParentDirectories()
+                        saveTo.writeBytes(screenshot.data)
+                        logger.info { "take_screenshot: saved to '$saveTo'" }
+                        "Screenshot saved to: $saveTo"
+                    } catch (e: Exception) {
```
AbdullinAM: Do we really need to have this nested try/catch? If we're worried about the IOExceptions, why not handle them in the top level try/catch? This file is already quite hard to read, so I would attempt to minimise the unnecessary complications

AbdullinAM (approved):

hot-reload-mcp/src/main/kotlin/org/jetbrains/compose/reload/mcp/McpServer.kt
pjBooms: removed

PR #556, opened 2026-06-18
[devtools] CMP-10310: Fix hot reload panel non-interactive in multi-window mode

pjBooms:
The sidecar windows were created as DocumentModal dialogs, which blocks all
other dialogs in the same document. Since all owner-less sidecars share a
single hidden AWT owner, they were in the same document and only the last-shown
window remained interactive.

Switch to Modeless dialog modality to allow all sidecar windows to be
interactive simultaneously.

Fixes: [CMP-10310](https://youtrack.jetbrains.com/issue/CMP-10310)


AbdullinAM (approved):
Nice find!

PR #557, opened 2026-06-23
[devtools] Fix reload state hanging indefinitely in continuous/auto mode

pjBooms:
In continuous `--auto` mode 'buildAndReload' launches a long-lived Gradle daemon
that never returns, blocking the recompiler loop so RecompileRequests were
never answered and ReloadState stayed 'Reloading' forever.

Run the continuous build once in a background subtask and respond each
RecompileRequest immediately.

Fixes: [CMP-10309](https://youtrack.jetbrains.com/issue/CMP-10309)

### DIFF
--- a/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/gradle/GradleRecompilerExtension.kt
+++ b/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/gradle/GradleRecompilerExtension.kt
@@ -42,7 +42,7 @@ internal class GradleRecompilerExtension : RecompilerExtension {
         }
 
         /* Side Effect */
-        if (HotReloadEnvironment.gradleWarmupEnabled || HotReloadEnvironment.gradleBuildContinuous) {
+        if (HotReloadEnvironment.gradleWarmupEnabled) {
             OrchestrationMessage.RecompileRequest().sendAsync()
         }
 
--- a/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/recompiler.kt
+++ b/hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/recompiler.kt
@@ -12,7 +12,9 @@ import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.flow.filterIsInstance
 import org.jetbrains.compose.devtools.api.RecompilerExtension
 import org.jetbrains.compose.reload.core.Environment
+import org.jetbrains.compose.reload.core.ExitCode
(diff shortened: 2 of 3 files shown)


hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/gradle/GradleRecompilerExtension.kt
```diff
@@ -42,7 +42,7 @@ internal class GradleRecompilerExtension : RecompilerExtension {
         }
 
         /* Side Effect */
-        if (HotReloadEnvironment.gradleWarmupEnabled || HotReloadEnvironment.gradleBuildContinuous) {
+        if (HotReloadEnvironment.gradleWarmupEnabled) {
```
AbdullinAM, line 45: Why do we change the behaviour here?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/recompiler.kt
```diff
@@ -57,38 +59,43 @@ internal fun launchRecompiler(): Future<Unit> = launchTask("Recompiler", recompi
...
-                    RecompileResult(request.messageId, exitCode.value?.code).send()
-                }
+        if (exitCode.isFailure()) logger.error("BuildSystem: '${recompiler.name}' failed", exitCode.exception)
+        else logger.debug("BuildSystem: '${recompiler.name}' finished w/ exitCode=${exitCode}")
+        return exitCode
+    }
+
+    val requests = messages.consumeAsFlow().filterIsInstance<RecompileRequest>()
+
+    if (HotReloadEnvironment.gradleBuildContinuous) {
```
AbdullinAM, line 85: Why do we need to change the behaviour of the recompiler at all here? Do I understand correctly that the original behaviour also just returned `RecompileResult` instantly?

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/gradle/GradleRecompilerExtension.kt
pjBooms: Because now we launch recompiler unconditionally for `buildContinous`, previously we launched it only for first `RecompileRequest` that was initiated here and has never returned (that's why it was the hang)

hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/recompiler.kt
pjBooms: No, it has never returned with any result. That was the root cause of the hang

PR #560, opened 2026-06-29
[mcpserver] Remove window title from screenshot.

pjBooms:
To match future screenshot capture implementation in Compose.

Window title is still available by `list_windows` tool.

### DIFF
--- a/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/screenshotHandler.kt
+++ b/hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/screenshotHandler.kt
@@ -53,14 +53,19 @@ internal fun handleScreenshotRequest(request: ScreenshotRequest, window: Window,
 /**
  * Captures the window content using [Robot.createScreenCapture].
  *
- * It captures the window title in addition to Compose content,
- * so AI agent can analyze and understand the full context of the window.
+ * The window decorations (title bar, borders) are excluded so only the Compose content is captured.
  */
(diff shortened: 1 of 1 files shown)


AbdullinAM (approved):

m-sasha:
Looks good, but a few points to think about.

- Does this work correctly with https://jetbrains.github.io/JetBrainsRuntimeApi/jetbrains.runtime.api/com/jetbrains/WindowDecorations.html ?
- Consider adding an `includeWindowTitle: Boolean = false` flag in case the agent wants the title anyway, in case it's needed after all. For example, on Windows the title can include a developer-settable icon.

pjBooms:
>Consider adding an includeWindowTitle: Boolean = false flag in case the agent wants the title anyway, in case it's needed after all. For example, on Windows the title can include a developer-settable icon.

I guess, we can add it later on a request if we would have a real world use case for this

pjBooms:
> Does this work correctly with https://jetbrains.github.io/JetBrainsRuntimeApi/jetbrains.runtime.api/com/jetbrains/WindowDecorations.html ?

As this API allows to mix title with the Compose content, we can't strip, for instance, system icons for minimizing/maximazing the window from the final screenshot. However, I believe it is quite a niche use case that we can neglect.

m-sasha (approved):

m-sasha:
Probably worthwhile to "tell" the agent that the window screenshot will not include the system title bar.

pjBooms:
> Probably worthwhile to "tell" the agent that the window screenshot will not include the system title bar.

Makes sense. Please take a look at https://github.com/JetBrains/compose-hot-reload/pull/561

PR #565, opened 2026-07-14
[runtime] Add hard-reset composition recovery under a runtime flag

pjBooms:
Hard reset is required when Compose Runtime cannot recover from corrupted state.
See [CMP-10459](https://youtrack.jetbrains.com/issue/CMP-10459), [CMP-10471](https://youtrack.jetbrains.com/issue/CMP-10471).

Hard-reset is added under a runtime flag. 
Enabled by default because even when the fixes for above bugs are landed to AOSP, we cannot be 100% sure that there won't remain other bugs. Hard-reset is better/faster then restart but has unfortunate side-effect: window blinks because it is also recreated.

### DIFF
--- a/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/compose.kt
+++ b/hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/compose.kt
@@ -5,6 +5,7 @@
 
 package org.jetbrains.compose.reload.agent
 
+import org.jetbrains.compose.reload.InternalHotReloadApi
 import org.jetbrains.compose.reload.analysis.ClassId
 import org.jetbrains.compose.reload.analysis.ComposeGroupKey
 import org.jetbrains.compose.reload.analysis.Ids
@@ -16,6 +17,7 @@ import org.jetbrains.compose.reload.core.warn
 import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.InvalidatedComposeGroupMessage
 import org.jetbrains.compose.reload.orchestration.OrchestrationMessage.InvalidatedComposeGroupMessage.DirtyScope
 import java.lang.instrument.ClassFileTransformer
+import java.lang.reflect.InvocationTargetException
 import java.lang.instrument.Instrumentation
 import java.security.ProtectionDomain
 import java.util.WeakHashMap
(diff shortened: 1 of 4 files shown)


hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/compose.kt
```diff
@@ -138,3 +139,49 @@ private fun invalidateGroupsWithKey(loader: ClassLoader, key: ComposeGroupKey) {
...
+
+/**
+ * Forces a full composition teardown and rebuild ("hard" reload).
+ *
+ * Unlike [invalidateGroupsWithKey] — which only invalidates changed groups and, crucially, *skips*
+ * any recomposer whose current error is non-recoverable — this disposes every running recomposer's
+ * compositions and recomposes them from scratch (Compose's 'saveStateAndDisposeForHotReload' followed
+ * by 'loadStateAndComposeForHotReload', which resets all error states).
+ */
+fun reloadCompositionState() {
```
AbdullinAM, line 153: I would at least recommend making this an internal API

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/resetComposition.kt
```diff
@@ -6,7 +6,26 @@
...
+ *    save/load ('reloadCompositionState'), which resets error states and
+ *    recreates the scene/window. Required on Compose runtimes that cannot recover from
+ *    a subcomposition error (See CMP-10459, CMP-10471).
+ *  - light reset: bump 'hotReloadState.key', which only disposes and recomposes the entry point's
+ *    subtree — no window recreation. Sufficient when Compose runtime can recover
+ *    from subcomposition errors.
+ */
 internal fun resetComposition() {
-    hotReloadState.update { state -> state.copy(key = state.key + 1) }
+    if (HotReloadEnvironment.compositionHardResetEnabled) {
```
AbdullinAM, line 26: Does it make sense to have this as a per-execution flag? Maybe we can have two different commands: reset UI and reset window (names could be better)?

AbdullinAM (changes requested):

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/compose.kt
pjBooms: The usage is in another module (`hot-reload-runtime-jvm`) which prevents it declaring `internal`. I can move it there. I declared it here (`compose.kt`) because it is similar to `invalidateGroupsWithKey` also declared here. WDYT is it better to move?

AbdullinAM: We can still use `@InternalHotReloadApi` even though it is used in another module

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/resetComposition.kt
pjBooms: I thought about it and decided that it complicates end-users usage without a valuable benefit. 
Really, an end-user will unlikely use the "Clear UI" button until something goes wrong. And if something goes wrong, the most probably he would want to repair it fast and reliably. And if we will have two buttons what button should he click? Why they are two? (btw, we had two buttons and still have two requests artifacts from that).  We already have two ways to recover: restart and clear ui, the third one would confuse users, imho.

Another legal scenario however, if a user wants  to "clear ui" on purpose -- to test the app from zero state. I am not sure how often the scenario is, but even for it bothering user to choose is not necessary, imho.

hot-reload-agent/src/main/kotlin/org/jetbrains/compose/reload/agent/compose.kt
pjBooms: Valid. Fixed.

AbdullinAM (approved):

PR #570, opened 2026-07-31
Make Maven Central publish resilient to "already published" failures

pjBooms:
- PublishToMavenCentralTask now treats a FAILED deployment whose errors are all "already exists" as success, so re-running after a `502 error` (where the upload actually succeeded) is idempotent

### DIFF
--- a/buildSrc/src/main/kotlin/org/jetbrains/compose/reload/build/tasks/PublishToMavenCentralTask.kt
+++ b/buildSrc/src/main/kotlin/org/jetbrains/compose/reload/build/tasks/PublishToMavenCentralTask.kt
@@ -23,6 +23,10 @@ import io.ktor.http.HttpHeaders
 import io.ktor.http.HttpStatusCode
 import io.ktor.http.headers
 import kotlinx.coroutines.runBlocking
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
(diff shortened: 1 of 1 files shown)


.teamcity/builds/PublishRelease.kt
```diff
@@ -133,8 +134,20 @@ object PublishToGradlePluginPortal : BuildType({
 
     dependencies {
         snapshot(PublishToMavenCentral) {
-            onDependencyFailure = FailureAction.FAIL_TO_START
```
AbdullinAM, line 136: I don't really see the point of this change and I think it can only create possibly unwanted situations where we publish something to gradle even though maven publication failed. Not because of the infra issues, but because of something actually wrong with the build. In that case, we will have to update the version before pushing a fix.

In my opinion, maven publication should always be the main indicator of the release publication, because if we published something there --- we can already use CHR (gradle plugin is also available from maven central).


I don't think we should treat Maven Central publication as a source of flakiness, if something happened with the build --- it **is** important and we **should** stop everything to check what has happened. IMO, adding a single workaround for 'already published' should be enough

.teamcity/builds/PublishRelease.kt
```diff
@@ -133,8 +134,20 @@ object PublishToGradlePluginPortal : BuildType({
...
+             * two deployments any tighter than ordering:
+             *  - reuseBuilds = ANY  -> reuse the already-finished Maven Central build instead of launching a
+             *    fresh deploy. This is important because the Central Portal upload is not idempotent: once a
+             *    version bundle is deployed, re-running 'deployMavenCentralDeployBundle' fails (e.g. "200 OK",
+             *    or a "502 Bad Gateway" that still published). Re-running this leg must not re-trigger it.
+             *  - onDependencyFailure/Cancel = IGNORE -> a flaky Maven Central result (the deploy failing even
+             *    though the artifacts actually made it to Maven Central) must not block publishing the Gradle
+             *    plugin, which builds its own artifacts from source and consumes nothing from that build.
+             */
+            reuseBuilds = ReuseBuilds.ANY
```
AbdullinAM: I don't think we should ever reuse failed builds. Again, you already added a fix for the 'already published' error, so the dependencies with that build should not be marked as 'failed'

AbdullinAM (changes requested):
I think this change creates an opportunity to accidentally publish  two different versions to Gradle and Maven. I think we should do everything to minimise the chances of that happening.

Change to the maven publication is OK, but I don't like the Gradle publication changes at all. Can you share your motivation in regards to those changes?

As an alternative thought:
If we have a (fast) single-time workaround to deal with those kind of maven publication failures, maybe we should monitor the situation for now and don't change the infra? Currently, we had a single failure in 1.5 years of biweekly publications, maybe we shouldn't throw away our safeguards just in case this happens again?

.teamcity/builds/PublishRelease.kt
pjBooms: Ok, agree. I removed the commit from the branch (only `already published` fix remained)

.teamcity/builds/PublishRelease.kt
pjBooms: I removed the commit from the branch (only already published fix remained)

AbdullinAM (approved):

PR #574, opened 2026-09-08
[runtime] CMP-10603: preserve window position and size after hard res…

pjBooms:
Fixes: [CMP-10603](https://youtrack.jetbrains.com/issue/CMP-10603)


hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -22,19 +26,67 @@ import org.jetbrains.compose.reload.core.trace
...
+     * The most recent bounds observed while the window was in its [WindowPlacement.Floating] state.
+     */
+    var normalBounds: Rectangle? = null
+}
+
+/**
+ * The top-level windows currently managed by [startWindowManager], keyed by their [WindowId] and kept
+ * in registration order.
+ */
+private val activeWindowsLock = ReentrantLock()
```
AbdullinAM, line 61: Wouldn't just a regular concurrent hash map work here?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -22,19 +26,67 @@ import org.jetbrains.compose.reload.core.trace
...
 import java.awt.event.WindowAdapter
 import java.awt.event.WindowEvent
 import java.beans.PropertyChangeListener
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 
 
 private val logger = createLogger()
 
+internal class ManagedWindow(val window: Window, val compositeKeyHashCode: CompositeKeyHashCode) {
```
AbdullinAM: I really don't like the idea of storing a global reference to an AWT window. I would at least make this a weak ref.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -22,19 +26,67 @@ import org.jetbrains.compose.reload.core.trace
...
+     */
+    var normalBounds: Rectangle? = null
+}
+
+/**
+ * The top-level windows currently managed by [startWindowManager], keyed by their [WindowId] and kept
+ * in registration order.
+ */
+private val activeWindowsLock = ReentrantLock()
+private val activeWindows = LinkedHashMap<WindowId, ManagedWindow>()
```
AbdullinAM: One thing that is not clear from the code for me right now: why can't we reuse the `WindowsState` for this? It already stores windows size and position. If you can match a newly created window to its previous "version" by windowId, you can find its size/position from windows state?

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
```diff
@@ -22,19 +26,67 @@ import org.jetbrains.compose.reload.core.trace
...
+    activeWindowsLock.withLock { activeWindows[windowId]?.normalBounds = window.bounds }
+}
+
+/**
+ * Registers [window] and returns a [WindowId] derived from its position in the composition
+ * ([compositeKeyHashCode]) plus an ordinal that differentiate windows sharing that position.
+ */
+private fun registerWindow(compositeKeyHashCode: CompositeKeyHashCode, window: Window): WindowId =
+    activeWindowsLock.withLock {
+        val ordinal = activeWindows.values.count { it.compositeKeyHashCode == compositeKeyHashCode }
```
AbdullinAM: Can't this produce conflicting ordinal values? In the situation where compositeKeyHashCode remains the same:
* open window A -> WindowId("$compositeKeyHashCode/0")
* open window B -> WindowId("$compositeKeyHashCode/1")
* close window A
* open window C -> WindowId("$compositeKeyHashCode/1")

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: We had to preserve order. Now I changed it to ArrayList

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: fixed

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: Good catch. So  it seems we can't re-create WindowId reliably, so I returned it back to Random UUID.

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
pjBooms: Your next comment proves that it is hard to keep WindowId after hard reset.  Moreover WindowState did not contain placement and normalBounds.  As they are needed for hard reset restore functionality only, I don't think it is good to expose them to WindowState

hot-reload-runtime-jvm/src/main/kotlin/org/jetbrains/compose/reload/jvm/window.kt
AbdullinAM: (haven't looked at all the changes yet) Do I understand correctly that we are relying on the fact that all windows are opened in the same order during reset? Wouldn't this result in a lot of flakiness on reset with multiple windows?

pjBooms: Not quite. We group by `compositeKeyHashCode` (the Window {} call-site identity) first , so Windows from distinct call sites match by that key alone, order-independent. Positional order is only used within one call site  and  it's deterministic compose order, afaiu.

AbdullinAM (approved):

PR #575, opened 2026-09-08
[devtools] Tooltips are displayed behind the error overlay fix

pjBooms:
Fixes: [CMP-10597](https://youtrack.jetbrains.com/issue/CMP-10597)


hot-reload-devtools/src/main/kotlin/org/jetbrains/compose/devtools/widgets/DtTooltip.kt
```diff
@@ -51,6 +52,12 @@ private val tooltipHideDelay = 100.milliseconds
 
 private val defaultTooltipOffset = DpOffset(1.dp, 1.dp)
 
+/** Number of tooltip windows currently visible. */
+private var visibleTooltipCount = 0
```
AbdullinAM: make this atomic?

AbdullinAM (approved):

